PR #26, opened 2026-04-10
Add CSV export for sa-bench rollup

weireweire:
## Summary
- add a human-readable benchmark-rollup.csv alongside the existing JSON rollup. So that we don't need to extract data from benchmark log manually.
- keep benchmark-rollup.json and downstream postprocess behavior unchanged. It use mean instead of median.
- derive total and decode GPU counts from the resolved runtime resources when available, with a filename fallback for total GPU count
- include median-based CSV metrics and optional P90 decode running request data for sglang disaggregated runs

## CSV Example
```csv
Config,Total GPU Count,Decode GPU Count,Concurrency,Total Token Throughput,Output Token Throughput,Median TTFT,Median TPOT,Median ITL,P90 Decode Running Requests,Output Token Throughput per User,Total Token Throughput per GPU
b200-fp8-glm5-8k1k_lowlat_4,64,56,512,46343.14,5148.868,67510.52,21.541,644.745,21,46.423,724.112
```

## Testing
- UV_CACHE_DIR=/tmp/uv-cache uv run pytest tests/test_sa_bench_rollup.py -q

### DIFF
--- a/src/srtctl/benchmarks/scripts/sa-bench/rollup.py
+++ b/src/srtctl/benchmarks/scripts/sa-bench/rollup.py
@@ -2,12 +2,37 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-"""Generate benchmark-rollup.json from sa-bench results."""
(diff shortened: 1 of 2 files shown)


weireweire:
@ishandhanani Could you help to review

YAMY1234 (approved):
LGTM

PR #27, opened 2026-04-10
TRTLLM nsys profiling harness + Dynamo OTEL tracing automation

karen-sy:
Adds harness for TRTLLM nsys profiling.
Registers two modes: `nsys-time` (time-based profiling boundaries) and `nsys` (TRTLLM iteration based boundaries), and adds the `trace-bench` benchmark as a profiling target option.

Example of `nsys-time` to add to recipe yaml:
```
profiling:
  type: nsys-time
  delay_secs: 400        # seconds from worker launch before capture starts (covers weight loading + autotuner)
  duration_secs: 30      # capture window length — same on all workers for aligned P/D profiles
  benchmark_duration_secs: 500  # must be > delay_secs + duration_secs
```

Example of `nsys`:
```
profiling:
  type: nsys  
  prefill:
    start_step: 10
    stop_step: 50
  decode:
    start_step: 1000
    stop_step: 1100
```

+ Also adds harness that enables OTEL tracing (https://github.com/ai-dynamo/dynamo/blob/main/docs/observability/logging.md) for Dynamo events observability. Add to yaml:
```
observability:
  enable_otel: true
  otel_endpoint: "http://ip.to.local.workstation:port"
````

### CODE REVIEW

File: src/srtctl/benchmarks/scripts/trace-replay/bench.sh [hunk 1]
```diff
@@ -3,148 +3,181 @@
...
-
-uv venv "${AIPERF_VENV}"
-uv pip install -p "${AIPERF_VENV}" "${AIPERF_SPEC}" tiktoken
-export PATH="${AIPERF_VENV}/bin:${PATH}"
-echo "aiperf $(aiperf --version 2>/dev/null || echo 'installed') in ${AIPERF_VENV}"
+# Force install aiperf to right branch
+echo "Installing aiperf..."
+uv venv /tmp/aiperf-venv
+source /tmp/aiperf-venv/bin/activate
+uv pip install "aiperf @ git+https://github.com/ai-dynamo/aiperf.git@b1dd72f2a1ca58b6e72bbaba66c1d76114b856a0" protobuf "transformers==4.57.3"
```
alec-flowers: Lets keep it where the user can specify in the config file the aiperf version and any extra packages to be installed. I think thats a better path

alec-flowers: It also then gets captured to be able to be replicated.

src/srtctl/benchmarks/scripts/trace-replay/bench.sh
```diff
@@ -3,148 +3,181 @@
...
 # Run small benchmark for warmup
-echo "Running warmup..."
-WARMUP_DIR="${ARTIFACT_DIR}/warmup"
-mkdir -p "${WARMUP_DIR}"
+echo ""
+echo "Running warmup benchmark..."
 aiperf profile \
     -m "${MODEL_NAME}" \
-    --tokenizer "${TOKENIZER_PATH}" \
+    --tokenizer "/model/" \
```
alec-flowers: Lets not hardcode this.

src/srtctl/core/slurm.py
```diff
@@ -265,12 +265,31 @@ def start_srun_process(
 
     logger.info("srun command: %s", shlex.join(srun_cmd))
 
+    # Build a clean environment: strip SLURM vars that conflict with nested srun.
+    # SLURM_TRES_PER_TASK conflicts with SLURM_CPUS_PER_TASK causing srun fatal errors.
+    # Task/CPU-bind vars cause parallel container imports (OOM) and wrong CPU affinity.
+    _SLURM_VARS_TO_UNSET = {
```
alec-flowers: Why did you need to add this?

src/srtctl/frontends/sglang.py
```diff
@@ -137,6 +138,8 @@ def start_frontends(
 
             # Build env vars
             env_to_set: dict[str, str] = {}
+            # Add OTEL env vars (before frontend env so OTEL_SERVICE_NAME can be overridden)
```
alec-flowers: Was SGLang supposed to be part of this refactor?

src/srtctl/benchmarks/base.py
```diff
@@ -59,6 +59,19 @@ def build_command(
         """
         ...
 
+    def get_extra_mounts(self, config: SrtConfig) -> dict[Path, Path]:
```
alec-flowers: What is this needed for?

alec-flowers (approved):

src/srtctl/benchmarks/base.py
karen-sy: Ah, was an artifact of the past. will rm

src/srtctl/frontends/sglang.py
karen-sy: ah thought this was Dynamo Sglang, and not Sgland standalone. will rm

src/srtctl/core/slurm.py
karen-sy: Did you not encounter pyxis errors without these about NUM_CPU_PER_TASK conflicting due to multiple places trying to set it? Maybe the Nvidia main doesn't have this problem, I'll try without

alec-flowers: I've never had to do this in my runs.

PR #28, opened 2026-04-10
Enable nsys profiling for vLLM

leo-cf-tian:
This change allows the nsys profiling configs in YAML to correctly work on vLLM. It relies on https://github.com/ai-dynamo/dynamo/pull/8068. The change does not work with torch profiles yet.


src/srtctl/core/schema.py
```diff
@@ -1106,6 +1107,43 @@ class SrtConfig:
...
+    def inject_vllm_profiling_args(self, data, **kwargs):
+        backend = data.get("backend")
+        if not isinstance(backend, dict) or backend.get("type") != "vllm":
+            return data
+
+        profiling = data.get("profiling")
+        if not profiling or not isinstance(profiling, dict):
+            return data
+
+        if profiling.get("type") != "nsys":
```
xinli-sw: seems like we are making zero assumptions on what data could be and constantly type-checking, is this the case?

xinli-sw: gemini told me this can work if 3.10

```
match obj:
    case {"a": {"b": {"c": "foo"}}}:
        print("Operate!")
    case _:
        print("Structure didn't match.")
```

add a test too?

leo-cf-tian: Some of this was written by Claude, so some of the type checking is probably extraneous but I just left it in for cases where the data is malformed. I suppose that it would probably be a better idea to use an easier method or do away with it all together.

xinli-sw:
closing, @hjjq  can you open a new PR with your local branch?

PR #12, opened 2026-04-07
Add lm-eval benchmark runner for InferenceX evals

Oseltamivir:
## Summary
Add InferenceX multi-node eval support through an `lm-eval` benchmark runner and eval-only orchestration path. Lets InferenceX run accuracy-only jobs against existing srt-slurm multi-node disaggregated recipes without running the throughput benchmark stage.

Copied from https://github.com/ishandhanani/srt-slurm/pull/245

## How
- Add an `lm-eval` benchmark runner that sources InferenceX's `benchmarks/benchmark_lib.sh` from a mounted `/infmax-workspace`.
- Mount `INFMAX_WORKSPACE` into the container as `/infmax-workspace` when provided.
- Add `EVAL_ONLY=true` handling in `do_sweep.py` so eval-only jobs start infra/workers/frontend, run
  the full model health check, skip throughput, and launch `lm-eval` directly.
- Keep `RUN_EVAL=true` behavior as a post-benchmark eval path for normal throughput jobs.
- Pass model/framework/topology metadata into the eval container, including served `MODEL_NAME`, prefill/decode TP/EP/DPA/worker counts, sequence length, precision, runner type, and eval concurrency. 
- Map srt-slurm `PREFILL_DP_ATTN` / `DECODE_DP_ATTN` env vars to the InferenceX `PREFILL_DP_ATTENTION` /`DECODE_DP_ATTENTION` names expected by `append_lm_eval_summary`.
- Copy eval outputs (`meta_env.json`, `results*.json`, `sample*.jsonl`) into `/logs/eval_results/` for launcher-side artifact pickup.
- Preserve partial eval artifacts on lm-eval failure while still returning the original eval failure
  code.
- Document the InferenceX lm-eval integration in `docs/accuracy.md`.

## What

For `EVAL_ONLY=true`:
- srt-slurm still starts the normal deployment topology.
- The throughput benchmark runner is skipped.
- `wait_for_model()` verifies the configured prefill/decode or aggregated worker counts.
- `lm-eval` runs against the OpenAI-compatible endpoint.
- Eval failure is fatal.
- Low score leads to failure

For `RUN_EVAL=true` without `EVAL_ONLY=true`:
- The normal benchmark runs first.
- `lm-eval` runs as a post-step if throughput succeeds.
- Eval failure is non-fatal to the benchmark result.
- Low score leads to failure

### Validation run
https://github.com/SemiAnalysisAI/InferenceX/actions/runs/24059388771

### InferenceX PR
[...]

PR #41, opened 2026-04-17
Add lm-eval benchmark runner for evals

Oseltamivir:
## Summary
Add InferenceX multi-node eval support through an `lm-eval` benchmark runner and eval-only orchestration path. Lets InferenceX run accuracy-only jobs against existing srt-slurm multi-node disaggregated recipes without running the throughput benchmark stage.

Copied from https://github.com/ishandhanani/srt-slurm/pull/245

## How
- Add an `lm-eval` benchmark runner that sources InferenceX's `benchmarks/benchmark_lib.sh` from a mounted `/infmax-workspace`.
- Mount `INFMAX_WORKSPACE` into the container as `/infmax-workspace` when provided.
- Add `EVAL_ONLY=true` handling in `do_sweep.py` so eval-only jobs start infra/workers/frontend, run
  the full model health check, skip throughput, and launch `lm-eval` directly.
- Keep `RUN_EVAL=true` behavior as a post-benchmark eval path for normal throughput jobs.
- Pass model/framework/topology metadata into the eval container, including served `MODEL_NAME`, prefill/decode TP/EP/DPA/worker counts, sequence length, precision, runner type, and eval concurrency. 
- Map srt-slurm `PREFILL_DP_ATTN` / `DECODE_DP_ATTN` env vars to the InferenceX `PREFILL_DP_ATTENTION` /`DECODE_DP_ATTENTION` names expected by `append_lm_eval_summary`.
- Copy eval outputs (`meta_env.json`, `results*.json`, `sample*.jsonl`) into `/logs/eval_results/` for launcher-side artifact pickup.
- Preserve partial eval artifacts on lm-eval failure while still returning the original eval failure
  code.
- Document the InferenceX lm-eval integration in `docs/accuracy.md`.

## What

For `EVAL_ONLY=true`:
- srt-slurm still starts the normal deployment topology.
- The throughput benchmark runner is skipped.
- `wait_for_model()` verifies the configured prefill/decode or aggregated worker counts.
- `lm-eval` runs against the OpenAI-compatible endpoint.
- Eval failure is fatal.
- Low score leads to failure

For `RUN_EVAL=true` without `EVAL_ONLY=true`:
- The normal benchmark runs first.
- `lm-eval` runs as a post-step if throughput succeeds.
- Eval failure is non-fatal to the benchmark result.
- Low score leads to failure

### Validation run
https://github.com/SemiAnalysisAI/InferenceX/actions/runs/24059388771

### InferenceX PR
https://github.com/SemiAnalysisAI/InferenceX/pull/1000

### DIFF
--- a/docs/accuracy.md
+++ b/docs/accuracy.md
@@ -1,6 +1,6 @@
 # Accuracy Benchmarks
 
-In srt-slurm, users can run different accuracy benchmarks by setting the benchmark section in the config yaml file. Supported benchmarks include `mmlu`, `gpqa` and `longbenchv2`.
+In srt-slurm, users can run different accuracy benchmarks by setting the benchmark section in the config yaml file. Supported benchmarks include `mmlu`, `gpqa`, `longbenchv2`, and `lm-eval`.
 
 ## Table of Contents
 
@@ -14,6 +14,7 @@ In srt-slurm, users can run different accuracy benchmarks by setting the benchma
   - [Example: Quick Validation](#example-quick-validation)
   - [Output](#output)
   - [Important Notes](#important-notes)
+- [lm-eval](#lm-eval)
 
 ---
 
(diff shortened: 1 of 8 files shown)


Oseltamivir:
@xinli-sw 

Hi, made the changes requested in #12 , more general now and not InferenceX-specific. Tested with InferenceX still https://github.com/SemiAnalysisAI/InferenceX/actions/runs/24812985409

ishandhanani:
Closing: `main` has an `lm-eval` benchmark runner (`benchmark.type: lm-eval`) and the eval-only path. Thank you for the original.

PR #67, opened 2026-04-24
vLLM DSv4 GB200 PD 042326

ywang96:

PR #71, opened 2026-04-24
Combine DeepSeek V4 recipe and tokenizer changes

alec-flowers:
## Summary
- combine the DeepSeek V4 GB200 recipe branch from PR #67 with the DeepSeek V4 tokenizer support from PR #68
- set the DeepSeek V4 SA-Bench recipes to use `custom_tokenizer: "deepseek_v4"`
- disable `use_chat_template` for those recipes because DeepSeek V4 does not ship a Jinja chat template

## Validation
- `UV_CACHE_DIR="$PWD/.uv-cache" uv run ruff check src/srtctl/benchmarks/scripts/sa-bench/backend_request_func.py`
- parsed all four `recipes/vllm/deepseek-v4-pro/8k1k/*.yaml` files with PyYAML


ywang96:
@alec-flowers FYI we've pushed another round of changes in my branch

PR #73, opened 2026-04-24
feat(sa-bench): add sglang DeepSeek-V4 tokenizer

YAMY1234:
## Summary

Adds a client-side tokenizer for **DeepSeek-V4-Pro** that matches sglang server behavior exactly, so `sa-bench`'s `input_tokens` lines up with the server's `#new-token`.

Uses the **existing** `module.path.ClassName` hook in `backend_request_func.get_tokenizer`; **no changes to sa-bench source** are required.

## Motivation

DeepSeek-V4 ships *no* Hugging Face chat template, so `tokenizer.apply_chat_template()` raises `ValueError`. sglang's server solves this internally by replacing the HF-template path with a hard-coded DSML encoder (`encoding_dsv4.encode_messages`) whenever `arch == "DeepseekV4ForCausalLM"` (see sgl-project/sglang#23600). Without a matching client-side encoder, sa-bench silently falls back to raw-text tokenization and its counts diverge from the server's — skewing ISL, TPOT, TTFT, and MTP accept-rate accounting.

## Design

Two files under `sa-bench/sa_bench_tokenizers/`:

- **`_sglang_encoding_dsv4.py`** — vendored byte-exact from sgl-project/sglang commit `f5d03db853862c8fb0e805df591bed883a71868b`. Apache-2.0 header + upstream SHA preserved, so the file can be dropped when sglang publishes an official client-side package.
- **`sglang_deepseek_v4.py`** — HF-compatible wrapper. Its `apply_chat_template` mirrors `serving_chat` *exactly*:
  1. Insert an empty `system` message if missing.
  2. `thinking_mode="chat"`, `reasoning_effort=None` (sglang defaults).
  3. Call the vendored `encode_messages(...)` to render the raw DSML string.
  4. `hf_tokenizer.encode(..., add_special_tokens=False)` (the encoder already emits `<｜begin▁of▁sentence｜>`).

Naming: the package is `sa_bench_tokenizers` (not `tokenizers`) to avoid shadowing the HuggingFace `tokenizers` top-level package when sa-bench is run with transformers/vllm installed.

## Usage (recipe)

```yaml
benchmark:
  type: "sa-bench"
  custom_tokenizer: "sa_bench_tokenizers.sglang_deepseek_v4.SGLangDeepseekV4Tokenizer"
  use_chat_template: true
  ...
```

## Files

- `src/srtctl/benchmarks/scripts/sa-bench/sa_bench_tokenizers/__init__.py`
[...]

PR #74, opened 2026-04-25
fix(sa-bench): auto-fallback when tokenizer has no chat template

YAMY1234:
## Summary

Some models (e.g. **DeepSeek-V4**) ship **no Hugging Face chat template**; their rendering happens entirely inside the engine via a hard-coded encoder. With the project-wide default `use_chat_template: true` introduced in #20, recipes that don't also set `custom_tokenizer` end up calling `tokenizer.apply_chat_template(...)` directly and crash with:

```
ValueError: Cannot use apply_chat_template() because tokenizer.chat_template is not set ...
```

Reported by @ishandhanani while running the dsv4 sa-bench recipes against srt-slurm without the new flags from #73.

## Fix

In `benchmark_serving.main()`, right after `get_tokenizer` returns, detect the no-template case:

- if `use_chat_template` is on **and** no `custom_tokenizer` plugin is configured **and** the tokenizer exposes neither `chat_template` nor `default_chat_template`,
- emit a loud `warnings.warn(...)` pointing at `custom_tokenizer` (e.g. `SGLangDeepseekV4Tokenizer` added in #73),
- and force `args.use_chat_template = False` so the run completes against the raw-text path.

This way:

- DSv4 recipe **without** the plugin -> no longer crashes; runs in raw-text mode (input_tokens may diverge from server #new-token, but the user is told why).
- DSv4 recipe **with** `custom_tokenizer: sa_bench_tokenizers.sglang_deepseek_v4.SGLangDeepseekV4Tokenizer` -> unchanged: real DSML rendering, exact parity with sglang server.
- Models with HF chat templates (DSR1, K2, Qwen, GLM, ...) -> unchanged.

## Test plan

- [x] `python3 -m py_compile benchmark_serving.py`
- [ ] Re-run dsv4 recipe without `custom_tokenizer`/`use_chat_template` overrides; confirm warning + successful sa-bench completion.
- [ ] Re-run an existing dsv4 recipe (e.g. `recipes/gb300-fp4/1k1k-dsv4/agg-low-latency-chat.yaml`) with `custom_tokenizer` set; confirm no behavior change vs. main.
- [ ] Re-run a non-DSv4 recipe (DSR1 / Kimi-K2.6) and confirm chat template path is unaffected.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/srtctl/benchmarks/scripts/sa-bench/benchmark_serving.py
+++ b/src/srtctl/benchmarks/scripts/sa-bench/benchmark_serving.py
@@ -840,6 +840,29 @@ def main(args: argparse.Namespace):
         custom_tokenizer=args.custom_tokenizer,
     )
 
+    # Some models (e.g. DeepSeek-V4) ship NO Hugging Face chat template; the
+    # server-side rendering happens entirely inside the engine. If a user runs
+    # such a model without supplying a `custom_tokenizer` plugin, the default
+    # `use_chat_template=True` would cause `tokenizer.apply_chat_template(...)`
+    # to raise. Auto-fallback to raw-text mode and warn loudly so the run does
+    # not silently break.
+    if args.use_chat_template and not args.custom_tokenizer:
+        has_template = bool(getattr(tokenizer, "chat_template", None)) or bool(
+            getattr(tokenizer, "default_chat_template", None)
+        )
+        if not has_template:
+            warnings.warn(
+                f"Tokenizer for '{tokenizer_id}' has no chat_template and no "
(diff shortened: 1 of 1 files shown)


YAMY1234:
Superseded by #76, which takes a fail-fast approach with an actionable error instead of a silent auto-fallback. Auto-fallback risks producing benchmark numbers from a different code path than the recipe intends (the client/server token counts can diverge, as this PR's own warning notes). #76 also fixes a separate silent bug where bench.sh warmup was missing CHAT_TEMPLATE_ARGS.

PR #90, opened 2026-04-27
feat(vllm): add per-node DP launch mode

esmeetu:
## Summary

- Add opt-in `backend.dp_launch_mode: per_node` for vLLM data-parallel endpoints.
- Launch one process per physical node with the full local GPU set while preserving the existing `per_gpu` default.
- Derive `--data-parallel-size-local`, `--data-parallel-start-rank`, `--data-parallel-address`, and DP-RPC settings from the allocated topology.
- Reserve the per-local-rank KV-event port range and validate that configured global DP size matches allocated GPUs.

Per-role mixed launch overrides were split into the separate stacked follow-up #271.

## Motivation

The existing per-GPU launcher restricts every process to one GPU. DeepGEMM MegaMoE and CUDA symmetric-memory paths can require node-local DP ranks to share one process and CUDA namespace. The opt-in per-node layout provides that namespace without changing existing recipes.

```yaml
backend:
  type: vllm
  dp_launch_mode: per_node
  vllm_config:
    prefill:
      data-parallel-size: 8
      data-parallel-hybrid-lb: true
    decode:
      data-parallel-size: 16
      data-parallel-hybrid-lb: true
```

## Validation

- `make check`
- Ruff and formatting pass.
- Full test suite: `865 passed, 2 skipped, 6 deselected`.
- DeepSeek V4 Pro AgentX job `9700` validated uniform per-node launch for `3xPDEP8 + 1xDDEP16` with decode AMXF4 MegaMoE: `26.85 req/s`, `91,193 TPGS` on 40 GPUs, p50 TTFT `8.24s`, and p50 TPOT `58.73ms`.

### DIFF
--- a/docs/config-reference.md
+++ b/docs/config-reference.md
@@ -422,6 +422,36 @@ Each worker leader gets a globally unique port starting at 5550:
 | decode_0  | 5552 |
 | decode_1  | 5553 |
 
+### vLLM DP launch mode
+
+vLLM data-parallel endpoints use one process per GPU by default. Set
+`dp_launch_mode: per_node` to launch one process per node and let vLLM
+manage the local DP ranks in a shared CUDA namespace:
+
+```yaml
+backend:
+  type: vllm
+  dp_launch_mode: per_node
+  vllm_config:
+    prefill:
+      data-parallel-size: 8
+      data-parallel-hybrid-lb: true
+    decode:
+      data-parallel-size: 16
+      data-parallel-hybrid-lb: true
+```
+
+| Value      | Process layout                                      |
+| ---------- | --------------------------------------------------- |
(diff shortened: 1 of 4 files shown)


alec-flowers:
Is there a reason to mix the launch modes? Or is 1 way better than the other? If we have a reason to choose one way then I would prefer to be opinionated.

esmeetu:
Split as suggested: #90 now contains only the opt-in global `per_node` launch mode. The per-role mixed-DP overrides moved to the separate stacked draft #271. `per_node` is useful when node-local ranks need a shared CUDA namespace (for example MegaMoE/symmetric-memory paths); this does not assert a universal performance advantage, and the existing `per_gpu` behavior remains the default.

PR #114, opened 2026-04-29
multi process tokenizer in benchmark_serving

fzyzcjy:


weireweire:
should we target dsv4 branch？

src/srtctl/benchmarks/scripts/sa-bench/bench.sh
```diff
@@ -104,6 +104,10 @@ if [ "$DATASET_NAME" = "random" ]; then
         --random-input-len "$ISL"
         --random-output-len "$OSL"
         --random-range-ratio "${RANDOM_RANGE_RATIO}"
+        # Parallel random prompt generation. Default 48 saturates a 144-core
+        # GB300 host. Override via env: RANDOM_NUM_WORKERS=8 etc. 0 = auto =
+        # min(cpu_count, 8). 1 = serial (no multiprocessing).
+        --random-num-workers "${RANDOM_NUM_WORKERS:-48}"
```
weireweire, line 110: could we just use 0 as default so that other machine can also run. And also clear the GB300 related comments.

weireweire:
since the dsv4 branch already merged to main, could we target to main now?

PR #115, opened 2026-04-29
feat: live in-flight batch-metrics snapshotter (opt-in)

YAMY1234:
## Summary

Adds an **opt-in, zero-dependency-on-production-path** feature that spawns a daemon thread during the benchmark stage.  
Every N seconds it re-parses prefill/decode worker logs and atomically overwrites `<log_dir>/batch_metrics.png`, giving a near-real-time view of the run without any external monitoring stack.

`live_metrics` lives under the `telemetry` block in `srtslurm.yaml` — it is a lightweight complementary signal (log polling, no scraper container required) that fits naturally alongside the existing DCGM/node_exporter telemetry.

### Files changed

| File | Purpose |
|---|---|
| `src/srtctl/analysis/__init__.py` | New `analysis` subpackage |
| `src/srtctl/analysis/batch_log_parser.py` | Incremental, stateful SGLang log parser — no plotting, no aggregation |
| `src/srtctl/analysis/live_metrics.py` | Background snapshotter thread + matplotlib renderer |
| `src/srtctl/core/schema.py` | New `LiveMetricsConfig` dataclass nested under `TelemetryConfig` |
| `src/srtctl/cli/mixins/benchmark_stage.py` | 3-line hook to start/stop the snapshotter around the benchmark proc |

### How to enable

In `srtslurm.yaml`:

```yaml
telemetry:
  live_metrics:
    enabled: true
    interval_seconds: 60   # default
    downsample: 1          # keep every sample
```

When disabled (default), **zero code paths are touched** — `try_start_snapshotter` returns `None` immediately.

### Plot layout

7 rows × 2 columns, semantically paired (prefill left, decode right):

| Row | Prefill | Decode |
|---|---|---|
| 0 | input throughput (token/s) | gen throughput (token/s) |
| 1 | #new-seq | #running-req |
| 2 | #new-token | #full token |
| 3 | #cached-token | full token usage |
| 4 | #prealloc-req | #prealloc-req |
| 5 | #queue-req | #queue-req |
| 6 | #inflight-req | #transfer-req |

Worker labels are shortened to `p0`…`pN` / `d0`…`dN`. Colour map is `tab20` (supports up to 20 workers).

### Design notes

- Parser uses a `byte_offset` per file — safe to call repeatedly on growing logs.
- Snapshotter swallows its own exceptions so a plotting failure never kills the benchmark.
- Atomic write: `fig.savefig(tmp) → os.replace(tmp, dst)` — readers never see a torn file.
- `matplotlib` is an **optional** soft dependency; the snapshotter degrades gracefully if missing.
- A "final tick" always runs on shutdown so the last state is captured even if the benchmark crashes early.

## Test plan

- [x] Smoke-tested locally against CoreWeave `491-tom-radixcache` logs — plot generated correctly
- [ ] End-to-end live smoke test on CoreWeave cluster pending
- [ ] `enabled: false` (default) — confirm no behaviour change in existing benchmarks

### DIFF
--- a/src/srtctl/analysis/live_metrics.py
+++ b/src/srtctl/analysis/live_metrics.py
@@ -69,35 +69,54 @@ def _global_origin(prefill: Iterable[FileSeries], decode: Iterable[FileSeries]) 
     return first_seen
 
 
+# ---------------------------------------------------------------------------
+# Plot layout: (prefill_metric, decode_metric) row pairs.
+# Use None to leave a cell empty. Only metrics listed here are plotted;
+# all parsed metrics remain available in LogState for other consumers.
+# To add/remove/reorder rows, edit this list — no other code needs to
+# change.
+# ---------------------------------------------------------------------------
+_PLOT_ROWS: list[tuple[str | None, str | None]] = [
+    ("input throughput (token/s)", "gen throughput (token/s)"),
+    ("#new-seq",                   "#running-req"),
+    ("#new-token",                 "#full token"),
+    ("#cached-token",              "full token usage"),
+    ("#prealloc-req",              "#prealloc-req"),
+    ("#queue-req",                 "#queue-req"),
+    ("#inflight-req",              "#transfer-req"),
+]
+
+
 def _render_png(
     state: LogState,
     output_path: Path,
     title: str,
     downsample: int,
 ) -> None:
-    """Render per-worker time-series for every metric to a PNG.
+    """Render per-worker time-series to a PNG.
 
-    Layout: prefill metrics in column 0, decode metrics in column 1, one
-    metric per row. Each axis carries one line per worker file.
+    Layout is driven by ``_PLOT_ROWS``: each entry is a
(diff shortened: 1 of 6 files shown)


src/srtctl/analysis/live_metrics.py
```diff
@@ -0,0 +1,342 @@
...
+    except Exception as e:
+        logger.debug("Live metrics: failed to load cluster config: %s", e)
+        return None
+
+    cfg = (cluster_config or {}).get("reporting", {}).get("live_metrics") if cluster_config else None
+    if not cfg or not cfg.get("enabled"):
+        return None
+
+    try:
+        import matplotlib  # noqa: F401  -- imported here so we fail fast & loud
```
ishandhanani: just have this be a default in srtslurm

YAMY1234: Done — matplotlib is now a required srtctl dependency (commit d269374), and the runtime ImportError fallback here is dropped in 22af2f1.

PR #119, opened 2026-04-29
feat: Add live monitor to SRT-SLURM

leo-cf-tian:
The following PR adds an opt-in TUI to SRT-SLURM to simplify the process of monitoring sweeps.

The TUI consists of a live dashboard displaying all relevant info such as SLURM queue state, job lifecycle stage, worker readiness, and benchmark metrics, as well as being able to quickly check configs and logs related to each job. The docs have been updated to reflect this functionality.

Here is an example of what the dashboard looks like:
image

Along with a detailed view of a job:
image


ishandhanani:
wow

ishandhanani:
Do you want to ping me on slack?

PR #125, opened 2026-05-03
fix: preflight accepts hf:* model paths and Docker image URIs

Thunderbeee:
The runtime (RuntimeContext.from_config) already classifies model.path starting with "hf:" as a HuggingFace ID and any model.container not starting with "/" or "./" as a Docker URI to forward to ``srun --container-image``.  Preflight was strictly more conservative and rejected both, breaking AIB CI which has always passed ``hf:org/model`` and Docker URIs through to srtctl.  Add early-return branches in _preflight_model and _preflight_container that mirror the runtime classification — if runtime would accept it, preflight accepts it too.  The ":" guard on the container branch keeps preflight strict about typo'd local paths that aren't URI-shaped.

### DIFF
--- a/src/srtctl/cli/monitor.py
+++ b/src/srtctl/cli/monitor.py
@@ -623,7 +623,7 @@ def _detail_view(job_id: str, jobs: list[dict], state: _State, term_height: int 
     stage_color = job_info["stage_color"] if job_info else "dim"
 
     job_title = (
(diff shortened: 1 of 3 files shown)


YAMY1234:
https://gitlab-master.nvidia.com/compute/anna/infrastructure/AIB/-/pipelines/50098898 
Verified by this CI

YAMY1234 (approved):

PR #130, opened 2026-05-06
recipes: Add dsv4 8k1k High Throughput Config

weireweire:
## Summary
- Add a GB300 FP4 DeepSeek V4 Pro 8k1k disaggregated SGLang recipe as `recipes/dsv4-pro/sglang/gb300-fp4/8k1k.yaml`.
- Include zipped high-throughput variants for 1/4/8/10/14 prefill workers against a 4-node decode worker.
- Keep the benchmark commands aligned with each resource point, up to 163840 prompts at max concurrency 16384.

## Validation
- Parsed the YAML with `python3`/PyYAML.
- Checked the `zip_override_8k1k_hightpt` resource list and final benchmark command.
- Ran `git diff --check`.

### DIFF
--- a/recipes/dsv4-pro/sglang/gb300-fp4/8k1k.yaml
+++ b/recipes/dsv4-pro/sglang/gb300-fp4/8k1k.yaml
@@ -1,0 +1,283 @@
+base:
+  name: dsv4-pro-gb300-fp4_8k1k_disagg_dep16
+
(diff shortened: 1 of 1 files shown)


weireweire:
supersede by https://github.com/NVIDIA/srt-slurm/pull/173

PR #131, opened 2026-05-06
Add DeepSeek-V4-Pro 1k/1k MTP-off SA recipes for GB200 (TEP8 + DEP16)

nv-yna:
## Summary

12 srt-slurm recipes for the DeepSeek-V4-Pro 1k/1k gen-only MTP-off SA reproduction sweep on GB200. Engine configs are byte-translated from the executed-sweep `ctx_config.yaml` + `gen_config.yaml` — the same runs that produced the published pareto curve. Two parallelism families:

- **TEP8** (ctx TP=8 attn_dp=true + decode TP=8 EP=8 attn_dp=false): c=1, 2, 4, 6, 12, 24, 128, 256
- **DEP16** (ctx TP=8 attn_dp=true + decode TP=16 EP=16 attn_dp=true): c=512, 1024, 1536, 2048

Beyond verbatim per-c engine values, four srt-slurm default-override fields are set explicitly to prevent silent runtime drift:

- `dynamo.install: false` — default `true` would pip-install Dynamo 0.8.0 over the recipe's container
- `frontend.enable_multiple_frontends: false` — default `true` would start nginx multi-frontend
- `benchmark.num_prompts_mult: 1` — default 10 would inflate prompt count by 10×
- `benchmark.use_chat_template: false` — default `true` would crash for DSV4-Pro (no registered `custom_tokenizer` adapter)

Plus three approved env-var deviations on workers (matching the executed sweep's per-c launch values):

- `UCX_CUDA_IPC_ENABLE_MNNVL=n`
- `UCX_RNDV_SCHEME=put_zcopy`
- `TLLM_BENCHMARK_REQ_QUEUES_SIZE=<concurrency>` (decode worker only)

## Test plan

- [x] Sweep executed end-to-end and pareto curve published
- [x] `srtctl dry-run` validates representative recipes (tep8 c=1 4-node, dep16 c=2048 6-node)
- [ ] Rerun a subset from these committed recipes to confirm reproducibility

### DIFF
--- a/recipes/DeepSeekV4-Pro/disagg/trtllm_dynamo/gb200_nvfp4/ISL1K_OSL1K/DEP16/1k1k_c1024.yaml
+++ b/recipes/DeepSeekV4-Pro/disagg/trtllm_dynamo/gb200_nvfp4/ISL1K_OSL1K/DEP16/1k1k_c1024.yaml
@@ -1,0 +1,109 @@
+# DSv4-Pro 1k/1k MTP-off SA recipe — DEP16 c=1024
+# Engine fields byte-translated from the executed-sweep ctx_config.yaml + gen_config.yaml.
+# Additions on top of the verbatim engine config:
+#   - trust_remote_code: true (DSV4-Pro safety)
+#   - host_cache_size: 0 in tep8 prefill kv_cache_config
+#   - 3 allowed env-var deviations on workers (UCX_*, TLLM_BENCHMARK_REQ_QUEUES_SIZE)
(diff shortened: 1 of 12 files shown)


ishandhanani:
Closing with the 2.0 merge (#407): `recipes/` no longer exists in this repository. Recipes live in the downstream InferenceMAX tree (`benchmarks/multi_node/srt-slurm-recipes/`), and the in-repo corpus is the curated `examples/` matrix. `srtctl migrate -f <recipe> --in-place` rewrites a v1 recipe to the 2.0 layout. Please reopen these downstream if they are still current.

PR #133, opened 2026-05-06
Initial integration of aiperf into srt-slurm

FrankD412:
This PR introduces an initial `aiperf` benchmark mode which allows for the use of the basic knobs (ISL, OSL, random-ratio) and adds the ability to specify a version/branch of `aiperf` so that custom versions can be used.

Additionally, there is a helper script that prints out the results in a similar fashion to what the `sa-bench` does. Here's a working example (this isn't the end product just a sample):

```
Output throughput per user (tok/s):      15.56
---------------Time to First Token----------------
Mean TTFT (ms):                          9101.97
Median TTFT (ms):                        4754.62
P99 TTFT (ms):                           62502.95
-----Time per Output Token (excl. 1st token)------
Mean TPOT (ms):                          64.26
Median TPOT (ms):                        66.22
P99 TPOT (ms):                           76.23
---------------Inter-token Latency----------------
Mean ITL (ms):                           73.44
Median ITL (ms):                         N/A
P99 ITL (ms):                            N/A
----------------End-to-end Latency----------------
Mean E2EL (ms):                          72330.18
Median E2EL (ms):                        71550.52
P99 E2EL (ms):                           119205.44
==================================================
------------------Power Summary--------------------
Total GPU Power (W):                     16068.27
Total GPU Energy (J):                    12107097.48
Output Tokens per Joule (tok/J):         5.00
==================================================
2026-05-06 10:47:10 - Concurrency 6144 complete
```


ishandhanani:
We should just use the custom benchmark mode

FrankD412:
@ishandhanani -- based on our discussion, can we reopen this?

src/srtctl/core/telemetry.py
```diff
@@ -36,12 +36,19 @@ def generate_telemetry_config(
...
+    storage_path is the container-side absolute path the scraper writes
+    final parquets to. Must be a fresh nested path (e.g.
+    ``/logs/telemetry/2026-04-27/myrun_4845``) — the scraper rejects
+    pre-existing storage directories.
+    """
     dcgm_exporter = telemetry.dcgm_exporter
     node_exporter = telemetry.node_exporter
-    if dcgm_exporter is None or node_exporter is None:
-        raise ValueError("Telemetry exporters must be configured before generating telemetry config")
+    if dcgm_exporter is None:
```
ishandhanani, line 43: Why this

src/srtctl/core/config.py
```diff
@@ -133,6 +133,14 @@ def resolve_config_with_defaults(user_config: dict[str, Any], cluster_config: di
         model["container"] = resolved_container
         logger.debug(f"Resolved container alias '{container}' -> '{resolved_container}'")
 
+    # Apply default environment variables (cluster defaults, recipe overrides)
+    cluster_env = cluster_config.get("default_environment")
```
ishandhanani, line 156: can you explain why?

src/srtctl/core/telemetry.py
FrankD412: This introduces the ability to only stand up DCGM without node-exporter. It makes it so that all of the containers aren't required if you just want to run with DCGM.

src/srtctl/core/config.py
FrankD412: This introduces the ability to add a central set of environment variables. For `aiperf` (or even for the frameworks) that it could be useful to have environment variables that are consistent across all jobs. I can strip this out if you'd like.

src/srtctl/cli/mixins/telemetry_stage.py
```diff
@@ -108,25 +103,44 @@ def start_telemetry(self) -> list[ManagedProcess]:
...
+        config_path.write_text(
+            generate_telemetry_config(
+                processes=self.backend_processes,
+                frontend_topology=topology,
+                runtime=self.runtime,
+                telemetry=telemetry,
             )
         )
 
+        telemetry_dir = self.runtime.log_dir / telemetry.storage_subdir / self.runtime.job_id
```
FrankD412: This was added to mitigate this error that kept popping up:

```
Error: "Storage path '/logs/telemetry' is an existing directory. Please specify a nested path like 'YYYY-MM-DD/run-name' (e.g., '2024-12-05/my-run')."
```

ishandhanani:
Closing: sa-bench runs on aiperf on `main` (`benchmark.aiperf_package`, `aiperf_args`), which covers what this introduced. Thank you for the groundwork.

PR #135, opened 2026-05-07
Handle null telemetry in live metrics startup

weireweire:
when telemetry not added in srtslurm.yaml, it failed with `attributeError: 'NoneType' object has no attribute 'get'`


weireweire:
@YAMY1234

PR #140, opened 2026-05-08
Prefer container infra binaries and prewarm Pyxis images

nv-roverman:
## Summary
  - Prefer NATS and etcd binaries available inside the container.
  - Preserve support for explicitly mounted infra binaries.
  - Prewarm Pyxis images before launching distributed services and workers.

## Testing
  - `uv run pytest tests/test_benchmarks.py`


ishandhanani:
Closing with the 2.0 merge (#407): etcd and NATS are services, so a recipe that prefers the container's own binaries declares them with a `command`. Image prewarming belongs to the caching track and should be its own PR against `main`.

PR #147, opened 2026-05-11
Default SA-Bench random workers to auto

weireweire:
Use SA-Bench's existing auto worker default for random prompt generation instead of a GB300-tuned 48-worker default.

Validation:
- bash -n src/srtctl/benchmarks/scripts/sa-bench/bench.sh


weireweire:
could we review this?

PR #156, opened 2026-05-14
Centralize safe runtime port allocation

weireweire:
Summary:
- Revert per-job SGLang port jitter in favor of centralized low, non-ephemeral runtime port ranges.
- Add preflight warnings for planned ports inside the kernel ephemeral range.
- Adjust busy planned worker ports before launch and log the old and new values.
- Update backend/frontend/infra defaults and related unit tests.

Validation:
- python3 -m py_compile on changed Python files and targeted tests
- bash -n src/srtctl/benchmarks/scripts/profiling/profile.sh
- git diff --check

Not run:
- pytest/ruff, because uv/pytest/ruff are not installed in this local environment and the system Python is too old for package runtime imports.


weireweire:
tested on mars and sglang and passed.

PR #157, opened 2026-05-14
vllm: support mooncake_kv_store + expose mooncake_master metrics

esmeetu:
## Summary

Extends `mooncake_master` orchestration (previously SGLang-only) to the **vLLM** backend so vLLM workers can use the in-process `MooncakeStoreConnector` for cross-process KV sharing. Also wires the master's admin HTTP server (`/metrics`, `/health`, etc.) on a configurable port so ops can scrape Prometheus metrics.

## What's new

- **`VLLMMooncakeKVStoreConfig`** (`backends/vllm.py`): mirrors the SGLang `mooncake_kv_store:` block and adds a vLLM-specific `store_config:` section. srtslurm renders that section into the `MOONCAKE_CONFIG_PATH` JSON file vLLM's `MooncakeStoreConnector` reads on startup (`MooncakeStoreConfig.load_from_env()`); the `env:` map is injected on every vLLM worker for in-process `MC_*` knobs (e.g. `MC_ENABLE_DEST_DEVICE_AFFINITY`, `MC_STORE_CLIENT_METRIC`, `MC_TE_METRIC`).
- **Shared launch constants** moved to a new `backends/mooncake.py`; the SGLang module re-exports them so existing imports keep working.
- **`start_mooncake_master`** now fires for vLLM as well as SGLang, writes the per-job store JSON next to the other log artifacts, and stamps the master RPC + HTTP-metadata endpoints on each worker.
- **Admin HTTP server**: master srun now passes `--enable_metric_reporting=true --metrics_port=9003` (upstream default in `master.cpp`), and `start_mooncake_master` waits on the metrics port like it already waits on RPC + HTTP-metadata. The flag toggles a periodic stdout log thread only — `MasterAdminServer` listens unconditionally — the new `MOONCAKE_METRICS_PORT` constant's comment documents that to prevent future reordering. Exposes `/metrics`, `/metrics/summary`, `/health`, `/role`, `/ha_status`, `/leader`, `/query_key`.

## Test plan

- [x] `make check` (ruff + pytest, 732 passed / 2 skipped)
- [x] New cases in `tests/test_e2e.py` cover vLLM mooncake env injection and JSON rendering
- [x] New cases in `tests/test_dry_run.py` cover the new config surface in `srtctl dry-run` output


src/srtctl/backends/vllm.py
```diff
@@ -207,13 +326,12 @@ def endpoints_to_processes(
 
         if not has_dp_mode:
             # Standard TP mode: one process per node
-            return endpoints_to_processes(endpoints, base_sys_port=base_sys_port, port_allocator=port_allocator)
+            return endpoints_to_processes(endpoints, base_sys_port=base_sys_port)
```
qiching: [do_sweep.py:108](https://github.com/NVIDIA/srt-slurm/blob/9cd8fe8e948bccefab2f83eece9b1dfbb806868c/src/srtctl/cli/do_sweep.py#L108) calls:

```python
allocator = NodePortAllocator.from_job_id(self.runtime.job_id)
return self.backend.endpoints_to_processes(self.endpoints, port_allocator=allocator)
```
After this change, vLLM jobs will crash with TypeError, which also violates the `BackendProtocol` interface in `base.py` (which includes port_allocator in the signature), and loses the job ID port jitter.

If you want to refactor `port_allocator` handling, we can a serepate PR to do.

src/srtctl/backends/vllm.py
```diff
@@ -132,14 +201,65 @@ def get_process_environment(self, process: Process) -> dict[str, str]:
...
+        - VLLM_NIXL_SIDE_CHANNEL_HOST: Routable IP for NIXL side channel (not 0.0.0.0/localhost)
         """
+        from srtctl.core.slurm import get_hostname_ip
+
         env: dict[str, str] = {}
         if process.kv_events_port is not None:
             env["DYN_VLLM_KV_EVENT_PORT"] = str(process.kv_events_port)
         if process.nixl_port is not None:
             env["VLLM_NIXL_SIDE_CHANNEL_PORT"] = str(process.nixl_port)
+            env["VLLM_NIXL_SIDE_CHANNEL_HOST"] = get_hostname_ip(process.node)
```
qiching: could i ask is a behavioral change related to mooncake? it adds a new env var to every vLLM worker and introduces a `get_hostname_ip` import. if unrelated, could you split it into a separate PR with its own test and description?

src/srtctl/backends/vllm.py
```diff
@@ -24,14 +24,78 @@
...
+              global_segment_size: "100GB"
+              local_buffer_size: "4GB"
+              protocol: "rdma"
+              device_name: ""
+              # master_server_address: srtslurm auto-fills from infra IP
+    """
+
+    container: str | None = None
+    env: dict[str, str] = field(default_factory=dict)
+    store_config: dict[str, str] | None = None
```
qiching: will be better to consider use `dict[str, Any]` here?

esmeetu:
Thanks for the review @qiching — addressed all three in 8306191:

1. **`endpoints_to_processes` signature** — added `port_allocator: NodePortAllocator | None = None` to match `BackendProtocol`, and threaded it through to `topology.endpoints_to_processes(...)` (both the standard-TP and DP-mode paths) so per-job port jitter from `NodePortAllocator.from_job_id` applies uniformly to vLLM jobs as well.
2. **Unrelated NIXL env var** — removed `VLLM_NIXL_SIDE_CHANNEL_HOST` and the `get_hostname_ip` import; that change isn't mooncake-related and shouldn't have been in this PR.
3. **`store_config` typing** — widened to `dict[str, Any]` (both the field on `VLLMMooncakeKVStoreConfig` and the return / local in `build_mooncake_store_config`). Added a short comment explaining that vLLM's `MooncakeStoreConfig` is a mix of `str`, `int`, and human-readable size strings, so `dict[str, str]` would force users to quote numeric values unnecessarily.

`make check` clean (732 passed, 2 skipped).

src/srtctl/backends/vllm.py
```diff
@@ -142,6 +216,53 @@ def get_process_environment(self, process: Process) -> dict[str, str]:
...
+        from ``mooncake_kv_store.store_config`` when set; missing keys fall back
+        to defaults. ``master_server_address`` is always auto-filled from the
+        infra node IP (any user-provided value is overridden — the user can't
+        know the infra IP at config time).
+        """
+        user_cfg: dict[str, Any] = {}
+        if self.mooncake_kv_store is not None and self.mooncake_kv_store.store_config:
+            user_cfg = dict(self.mooncake_kv_store.store_config)
+
+        return {
```
qiching: could we not hardcore it? If vLLM upstream adds a new field to `MooncakeStoreConfig`, users will set it in YAML, it'll deserialize fine, but `build_mooncake_store_config` will swallow it and the JSON will never contain it.
like we can:
```python
defaults: dict[str, Any] = {
    "metadata_server": "P2PHANDSHAKE",
    "global_segment_size": "4GB",
    "local_buffer_size": "4GB",
    "protocol": "rdma",
    "device_name": "",
}
result = {**defaults, **user_cfg}
result["master_server_address"] = f"{infra_node_ip}:{MOONCAKE_MASTER_PORT}"
return result
```

how do you think?

CLAUDE.md
```diff
@@ -153,7 +153,7 @@ infra:
 
 When `mooncake_kv_store` is set under an SGLang backend, srtslurm:
```
qiching: this line still says "under an SGLang backend" but the PR extends `mooncake_kv_store` to vLLM as well. could we update `vllm` as well?

esmeetu:
@qiching It should be clean now.

qiching (approved):
LGTM. Good job. Thank you! @esmeetu

xinli-sw (approved):

PR #179, opened 2026-05-28
fix(slurm): use --key=value for srun options (Slurm 25.11 cpu-bind regression)

### LINKED ISSUE #127: cpu-bind command line issue
On newer SLURM clusters, I've encountered a cpu-bind error 

```
2026-05-05 08:00:06 [INFO] Polling http://nsc-svg-slurm-1-gpu-133:8000/health every 10.0s for 0 prefills and 1 decodes
srun.real: error: unrecognized --cpu-bind argument "--job-name"
srun.real: fatal: Failed to parse --cpu-bind= values.

```

 src/srtctl/backends/trtllm.py emits a space between the flags that caused an issue
```
    if cpu_bind:
        srun_cmd.extend(["--cpu-bind", cpu_bind])
```

Proposed fix: remove the space
```
if cpu_bind:
    srun_cmd.append(f"--cpu-bind={cpu_bind}")
```

shljessie:
**Problem**
Slurm 25.11.x has a regression where space-separated long options (--cpu-bind verbose,none) slurp the next argv token (e.g. --job-name=... from the sbatch allocation) as part of the value, failing with:

```
srun.real: error: unrecognized --cpu-bind argument "--job-name"
srun.real: fatal: Failed to parse --cpu-bind= values.
```
This breaks every TRT-LLM job (it's the only backend that sets cpu_bind="verbose,none" in backends/trtllm.py). I hit this during a gpt-oss-120b sweep — workers died in zero seconds with no log file ever created. Same bug independently reported in [#127](https://github.com/NVIDIA/srt-slurm/issues/127) by @faradawn.

**Fix**
Emit srun options as a single argv token (--key=value) instead of two (--key value) in src/srtctl/core/slurm.py. The = form is canonical Unix long-option syntax, parses unambiguously on all Slurm versions, and is already used elsewhere in the file (--container-image=, --mpi=, etc.).

Both code paths fixed:

- The dedicated cpu_bind parameter (used by TRT-LLM)
- The generic srun_options dict (any user routing cpu-bind through YAML hits the same crash)

Also updates show_config_details() in submit.py so srtctl dry-run shows what actually runs, and adds a regression test in tests/test_slurm.py.

Closes #127.

### DIFF
--- a/src/srtctl/cli/submit.py
+++ b/src/srtctl/cli/submit.py
@@ -276,8 +276,10 @@ def show_config_details(config: SrtConfig) -> None:
         console.print("[dim]No custom environment variables configured.[/]")
 
     # --- srun options ---
+    # Render with `=` to mirror what core/slurm.py emits at runtime; the
+    # space-separated form misparses `cpu-bind` on Slurm 25.11.x.
     if config.srun_options:
-        opts = " ".join(f"--{k} {v}" if v else f"--{k}" for k, v in config.srun_options.items())
+        opts = " ".join(f"--{k}={v}" if v else f"--{k}" for k, v in config.srun_options.items())
         console.print(f"[dim]srun options:[/] {opts}")
 
     show_extensions = (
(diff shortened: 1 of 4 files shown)


faradawn:
LGTM. I also had the same issue and had to modify the submit.py and other scripts.

Issue #186, opened 2026-05-29
Cannot Sweep EP values for vLLM


faradawn:
For vLLM, we sweep EP values with 

enable-expert-parallel: "{ep}"
sweep:
  ep: [true, false]

However, this results in an error 

__main__.py: error: unrecognized arguments: True

That came from this generated argument:

--enable-expert-parallel True

Is there a way to sweep EP value?

ishandhanani:
Closing as part of the post-2.0 issue sweep; the fix is in #293 (duplicates #187 and #213 were closed), which stays open and should be rebased onto `main` after #407.

PR #187, opened 2026-05-29
Support to EP sweeps for vLLM

faradawn:
Previously we couldn't sweep EP values [true, false] for vLLM. This PR added the support. 

Address issue: https://github.com/NVIDIA/srt-slurm/issues/186


ishandhanani:
Closing as a duplicate of #293 (same issue, #186); #293 carries the regression tests and stays open.

PR #199, opened 2026-06-05
Add numa memory bind in trtllm disagg by default

richardhuo-nv:
## Summary

- Adds `gpu_type` field to `RuntimeContext`, propagated from `config.resources.gpu_type`
- Prefixes TRTLLM worker commands with `numactl -m 0,1` when both conditions are met:
  - `gpu_type` is `"gb200"` or `"gb300"`
  - worker mode is `"prefill"` or `"decode"` (disaggregated only — skipped for aggregated)

## Motivation

GB200/GB300 NVLink-Switch systems have a split NUMA topology. Without explicit NUMA binding, TRTLLM worker processes may allocate memory across NUMA nodes, causing remote memory access latency. Binding to nodes 0 and 1 keeps allocations local to the correct socket.

This only applies to disaggregated prefill/decode workers — aggregated mode does not benefit from the same binding and is excluded.

## Changes

| File | Change |
|------|--------|
| `core/runtime.py` | Add `gpu_type: str = ""` field to `RuntimeContext`; populate from `config.resources.gpu_type` in both `from_config()` construction sites |
| `backends/trtllm.py` | Prepend `numactl -m 0,1` to the worker command when `gpu_type in ("gb200", "gb300")` and `mode in ("prefill", "decode")` |

### DIFF
--- a/src/srtctl/backends/trtllm.py
+++ b/src/srtctl/backends/trtllm.py
@@ -178,7 +178,12 @@ class TRTLLMProtocol:
         container_config_path = Path("/logs") / config_filename
         container_model_path = Path("/model")
 
+        numactl_prefix = (
+            ["numactl", "-m", "0,1"] if runtime.gpu_type in ("gb200", "gb300") and mode in ("prefill", "decode") else []
+        )
+
         cmd = [
+            *numactl_prefix,
             "trtllm-llmapi-launch",
             "python3",
             "-m",
--- a/src/srtctl/core/runtime.py
+++ b/src/srtctl/core/runtime.py
@@ -123,6 +123,9 @@ class RuntimeContext:
     # Request plane for dynamo workers
     request_plane: str = "nats"
 
+    # GPU type (e.g. "gb200", "gb300", "h100")
+    gpu_type: str = ""
+
     @classmethod
     def from_config(
         cls,
@@ -261,6 +264,7 @@ class RuntimeContext:
             environment=dict(config.environment),
             is_hf_model=is_hf_model,
             request_plane=config.dynamo.request_plane,
+            gpu_type=config.resources.gpu_type,
         )
 
         # Expand FormattablePath mounts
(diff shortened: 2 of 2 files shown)


qiching (commented):
Overall looks reasonable. A few things worth confirming before merge:

- numactl only wraps trtllm-llmapi-launch. The membind is inherited by locally-forked MPI ranks, but for cross-node TP/EP the remote ranks won't inherit it. Can you confirm with numactl will show in a worker log that binding lands on every rank?
- Please add a comment explaining the GB200/GB300 Grace NUMA layout behind 0,1, and confirm numactl is present in the TRTLLM image (otherwise the worker fails with command not found)?

richardhuo-nv:
> Overall looks reasonable. A few things worth confirming before merge:
> 
> * numactl only wraps trtllm-llmapi-launch. The membind is inherited by locally-forked MPI ranks, but for cross-node TP/EP the remote ranks won't inherit it. Can you confirm with numactl will show in a worker log that binding lands on every rank?
> * Please add a comment explaining the GB200/GB300 Grace NUMA layout behind 0,1, and confirm numactl is present in the TRTLLM image (otherwise the worker fails with command not found)?

yes, I can confirm both. The reason why I added it is because this becomes a standard in TRTLLM team's benchmark, 
see: https://gitlab-master.nvidia.com/ibc/bench-trtllm-disagg/-/blob/main/trtllm-disagg-benchmark/config.yaml?ref_type=heads#L10

This is the only way we match precisely with trtllm-serve's benchmark number.s

qiching:
> > Overall looks reasonable. A few things worth confirming before merge:
> > 
> > * numactl only wraps trtllm-llmapi-launch. The membind is inherited by locally-forked MPI ranks, but for cross-node TP/EP the remote ranks won't inherit it. Can you confirm with numactl will show in a worker log that binding lands on every rank?
> > * Please add a comment explaining the GB200/GB300 Grace NUMA layout behind 0,1, and confirm numactl is present in the TRTLLM image (otherwise the worker fails with command not found)?
> 
> yes, I can confirm both. The reason why I added it is because this becomes a standard in TRTLLM team's benchmark, see: https://gitlab-master.nvidia.com/ibc/bench-trtllm-disagg/-/blob/main/trtllm-disagg-benchmark/config.yaml?ref_type=heads#L10
> 
> This is the only way we match precisely with trtllm-serve's benchmark number.s

 makes sense. That addresses my main concerns.

qiching (approved):
LGTM

PR #213, opened 2026-06-10
Preserve sweep placeholder value types

### LINKED ISSUE #186: Cannot Sweep EP values for vLLM

fallintoplace:
## Summary

Preserve the original type when a sweep placeholder is the entire config value, instead of stringifying booleans and numbers. This lets vLLM boolean flags such as `enable-expert-parallel: "{ep}"` expand to real booleans, so `True` emits `--enable-expert-parallel` without an extra positional value and `False` omits the flag.

Fixes #186.

## Tests

- `uv run pytest tests/test_sweep.py tests/test_configs.py::TestVLLMDataParallelMode::test_dp_mode_command_includes_dp_flags -q`
- `uv run ruff check src/srtctl/core/sweep.py tests/test_sweep.py tests/test_configs.py`
- `uv run pytest tests -q`

### DIFF
--- a/src/srtctl/core/sweep.py
+++ b/src/srtctl/core/sweep.py
@@ -32,13 +32,12 @@ def expand_template(template: Any, values: dict[str, Any]) -> Any:
         result = template
         for key, value in values.items():
             placeholder = f"{{{key}}}"
+            if result == placeholder:
+                return value
             # Handle list values specially - convert to comma-separated string or keep as list
             if isinstance(value, list):
                 # For YAML lists, we want to keep them as lists, not convert to string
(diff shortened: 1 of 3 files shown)


ishandhanani:
Closing as a duplicate of #293 (same issue, #186); #293 carries the regression tests and stays open.

PR #218, opened 2026-06-14
gb300 mxfp4 EPLB: mount tmpfs /dev/shm size=100% for expert loading

xinli-sw:
EPLB loads all experts into host /dev/shm during moe_load_balancer finalize. With the small default container /dev/shm, large-MoE expert sets overflow it and prefill workers die with SIGBUS ("nonexistent physical address") in UCX, so they never register and the run times out at the health-check ceiling.

Mount a tmpfs at /dev/shm sized to 100% of host RAM on the 36 gb300_mxfp4 EPLB (eplb384) disagg recipes. The eplb0 recipes do not use shared-expert memory and are left unchanged.

Also let srtctl pass a "tmpfs:" extra_mount source through to pyxis verbatim instead of resolving it as a host path.


xinli-sw:
testing on q2 branch, low risk change, will revert if does not work.

PR #234, opened 2026-06-30
dynamo: remap container root on dynamo-install sruns via ENROOT_REMAP_ROOT

Ankur-singh:
## Problem

The `dsv4 fp4 gb300 dynamo-sglang` sweep failed 4/4 configs at "Installing dynamo from source" (exit 100). Failing run: [SemiAnalysisAI/InferenceX Actions run 27657243156](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/27657243156) (PR SemiAnalysisAI/InferenceX#1697) — e.g. `dsv4_1k1k fp4 gb300-nv dynamo-sglang | P(tp4/ep4) D(tp8/ep8) | disagg-true spec-mtp conc-8192`, which fails at the "Launch multi-node job script" step. Root cause (A/B-proven):

- A cold dynamo source build runs `apt-get install` / `pip install --break-system-packages` — which need **root**.
- srtctl launches its containers as the **unprivileged** user, so the install dies with `dpkg: requires superuser privilege` (the worker log is SIGKILL-truncated mid-install).
- Clean A/B in the same container as the real CI user: **without** root remap → exit 100; **with** it → RC 0.

## Fix

Inject `ENROOT_REMAP_ROOT=yes` (the enroot var that remaps the container user to root at container-creation time) **only on the sruns that install dynamo**, and a no-op everywhere else. Recipe authors set nothing.

The dynamo install (`DynamoConfig.get_install_commands()`) runs in exactly two places, both of which previously launched as the unprivileged user:

1. **Backend workers** — `worker_stage.py`, gated on `frontend.type == "dynamo" and dynamo.install` (both the SGLang per-node and TRTLLM/MPI launch paths).
2. **Dynamo frontend** — `frontends/dynamo.py`, gated on `dynamo.install`.

Everything else (etcd/nats, mooncake, model pre-download, post-eval, benchmark, telemetry, nginx, sglang frontend) is untouched.

## Why an env var, not the `--container-remap-root` pyxis flag

This was the key design decision. pyxis exposes a `--container-remap-root` srun flag that does the same remap, but we deliberately use the enroot env var instead:

| | `--container-remap-root` (pyxis flag) | `ENROOT_REMAP_ROOT=yes` (enroot env var) ✅ |
|---|---|---|
| Cluster doesn't support it | **srun hard-fails**: an unrecognized option makes srun print `unrecognized option` and **exit non-zero before the step launches** — it does not degrade to a no-op. Blindly adding it would break every dynamo srun on a cluster whose pyxis predates the flag. | **Graceful no-op**: srun never parses the var, so an unsupporting cluster simply ignores it (the container starts unprivileged as before) instead of failing the step. |
| Layer it talks to | pyxis (the SLURM SPANK plugin) — the scheduler's flag surface. | enroot directly, which actually performs the uid remap. enroot reads `ENROOT_REMAP_ROOT` (enroot.conf default `no`), and per the enroot docs *"environment variables take precedence over the configuration file."* |
| Scope | per-srun. | per-srun, and **does not** touch the cluster-wide enroot.conf default. |

"Don't let the srun fail" was the hard requirement, and only the env var satisfies it.

### Why `--export`, not the existing `env_to_set`

enroot reads `ENROOT_REMAP_ROOT` when it **creates** the container. srtctl's `env_to_set` becomes `export NAME=value` statements **inside** the container's `bash -c` wrapper — i.e. *after* the container already exists, far too late for the remap. So the var must reach the **task/step** environment, which we do with `--export=ALL,ENROOT_REMAP_ROOT=yes`:

- `ALL` preserves srun's normal full-environment propagation (its usual default), and we only add the one var on top — no change to non-dynamo sruns.
- It's visible in the logged srun command and in `srtctl dry-run`, so it's debuggable rather than hidden.

## Changes

- `core/slurm.py`: new `srun_export_env` param on `start_srun_process` → renders `--export=ALL,K=V,…`; `CONTAINER_REMAP_ROOT_EXPORT` constant.
- `cli/mixins/worker_stage.py`: `_installs_dynamo()` as the single source of truth (reused by the preamble gate and both worker srun calls).
- `frontends/dynamo.py`: inject on the frontend srun when `dynamo.install`.
- `cli/submit.py`: `dry-run` surfaces the injected export line.
- Tests: argv rendering (present/absent), worker + frontend wiring (dynamo / sglang / `install: false`), and dry-run visibility — 8 new tests.

## Testing

- New tests pass; full suite **751 passed** (the 2 unrelated failures are pre-existing and reproduce on `main`: local ANSI-color substring matching, and Python 3.14's `UnionType[...]` repr vs the repo's 3.10 target). `ruff check` + `ruff format` clean.
- **Cluster checks (recommended before merge):**
  1. Confirm pyxis forwards the var to enroot: `srun --jobid=$JOB_ID --overlap --container-image=$IMG --export=ALL,ENROOT_REMAP_ROOT=yes bash -c 'whoami'` → expect `root`.
  2. Re-run the previously failing `dsv4 fp4 gb300 dynamo-sglang` config (cold cache) → expect the source install completes (RC 0).

### DIFF
--- a/src/srtctl/cli/submit.py
+++ b/src/srtctl/cli/submit.py
@@ -305,6 +305,13 @@ def show_config_details(config: SrtConfig) -> None:
         opts = " ".join(f"--{k}={v}" if v else f"--{k}" for k, v in config.srun_options.items())
         console.print(f"[dim]srun options:[/] {opts}")
 
+    # Dynamo install runs apt-get/pip as root inside the container, so srtctl injects
+    # ENROOT_REMAP_ROOT=yes (via srun --export) on the worker + dynamo-frontend launches.
+    if config.frontend.type == "dynamo" and config.dynamo.install:
+        console.print(
+            "[dim]srun --export (dynamo install):[/] ALL,ENROOT_REMAP_ROOT=yes [dim](workers + dynamo frontend)[/]"
+        )
+
     show_extensions = (
         config.benchmark.type == "custom"
         or config.benchmark.container_image
--- a/src/srtctl/cli/submit.py
+++ b/src/srtctl/cli/submit.py
@@ -55,7 +55,7 @@ from srtctl.core.git_state import (
     write_git_state_snapshot,
 )
 from srtctl.core.lockfile import load_lockfile_fingerprints
-from srtctl.core.schema import SrtConfig
+from srtctl.core.schema import SrtConfig, installs_dynamo
 from srtctl.core.status import create_job_record
 from srtctl.core.validation import preflight_config_variants
 from srtctl.ports import MOONCAKE_MASTER_PORT
@@ -307,7 +307,7 @@ def show_config_details(config: SrtConfig) -> None:
 
     # Dynamo install runs apt-get/pip as root inside the container, so srtctl injects
     # ENROOT_REMAP_ROOT=yes (via srun --export) on the worker + dynamo-frontend launches.
-    if config.frontend.type == "dynamo" and config.dynamo.install:
+    if installs_dynamo(config):
         console.print(
             "[dim]srun --export (dynamo install):[/] ALL,ENROOT_REMAP_ROOT=yes [dim](workers + dynamo frontend)[/]"
         )
--- a/src/srtctl/cli/mixins/worker_stage.py
+++ b/src/srtctl/cli/mixins/worker_stage.py
@@ -15,7 +15,7 @@ from typing import TYPE_CHECKING, Any
 from srtctl.core.fingerprint import generate_capture_script
 from srtctl.core.processes import ManagedProcess, NamedProcesses
 from srtctl.core.schema import build_otel_env
-from srtctl.core.slurm import get_hostname_ip, start_srun_process
+from srtctl.core.slurm import CONTAINER_REMAP_ROOT_EXPORT, get_hostname_ip, start_srun_process
 from srtctl.ports import ETCD_CLIENT_PORT, KV_EVENTS_PORT_BASE, KVBM_ZMQ_PORT_BASE, NATS_PORT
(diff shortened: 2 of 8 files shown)


src/srtctl/cli/submit.py
```diff
@@ -305,6 +305,13 @@ def show_config_details(config: SrtConfig) -> None:
         opts = " ".join(f"--{k}={v}" if v else f"--{k}" for k, v in config.srun_options.items())
         console.print(f"[dim]srun options:[/] {opts}")
 
+    # Dynamo install runs apt-get/pip as root inside the container, so srtctl injects
+    # ENROOT_REMAP_ROOT=yes (via srun --export) on the worker + dynamo-frontend launches.
+    if config.frontend.type == "dynamo" and config.dynamo.install:
```
csahithi: Just a small design change. Approving the PR though, feel free to merge this PR without this suggestion for now
`_installs_dynamo()` can be reused here if it is created in schema.py
```
def installs_dynamo(config: "SrtConfig") -> bool:
    return config.frontend.type == "dynamo" and config.dynamo.install
```
and called as `installs_dynamo(self.config)`

csahithi (approved):

src/srtctl/cli/submit.py
Ankur-singh: Done in b213ff0 — extracted `installs_dynamo(config)` into `schema.py` and reused it in both `worker_stage` and the `submit.py` dry-run display (removed the local `_installs_dynamo`). Kept it as a module-level function (not a property) so it also works with the lightweight `SimpleNamespace` configs in the unit tests. Thanks for the review!

PR #235, opened 2026-07-01
feat(nginx): configurable session-affinity header (generalizes #233)

YAMY1234:
## Summary

Generalizes the nginx session-affinity approach from #233 so the hashed header is **configurable** via `frontend.nginx_session_affinity_header` (default `X-Dynamo-Session-ID`, which preserves #233's behavior).

**Why:** not every client sends `X-Dynamo-Session-ID`. aiperf's conversation-aware routing (`--use-dynamo-conv-aware-routing`) carries the session id in the request **body** (`nvext.session_control.session_id`) and in the **`X-Correlation-ID`** header — it never sets `X-Dynamo-Session-ID`. On an aiperf + body-session Dynamo stack, hashing `X-Dynamo-Session-ID` is therefore a no-op (empty header → falls back to `$request_id` → round-robin). Setting the header to `X-Correlation-ID` pins each conversation to one frontend for such clients.

## Changes
- `FrontendConfig`: add `nginx_session_affinity` (bool) + `nginx_session_affinity_header` (str, default `X-Dynamo-Session-ID`)
- frontend stage passes both into the nginx template
- `nginx.conf.j2`: when enabled, `map` the configured header into a hash key (empty → `$request_id`) and add `hash ... consistent` on the upstream

## Validation (GB300, Qwen3.5-397B-NVFP4, AgentX cc-traces, 3P1D-DEP8, conc 256, 5 frontends)

A/B differ only in `nginx_session_affinity` (B hashes `X-Correlation-ID`):

| Metric | multi-FE, affinity OFF | multi-FE, affinity ON | single-FE ref |
|---|--:|--:|--:|
| cache hit | 56.0% | 93.1% | 93.0% |
| TPS/GPU | 22,012 | 50,521 (2.29×) | 47,589 |
| TTFT p50 | 32.0 s | 7.0 s | — |

Pinning the conversation to one frontend recovers prefix cache to single-frontend level and 2.29× throughput. Rendered `nginx.conf` verified for both header values and the off case.

### DIFF
--- a/src/srtctl/cli/mixins/frontend_stage.py
+++ b/src/srtctl/cli/mixins/frontend_stage.py
@@ -181,6 +181,8 @@ class FrontendStageMixin:
             backend_port=topology.frontend_port,
             listen_port=topology.public_port,
             nginx_raise_ulimit=self.config.frontend.nginx_raise_ulimit,
+            nginx_session_affinity=self.config.frontend.nginx_session_affinity,
+            nginx_session_affinity_header=self.config.frontend.nginx_session_affinity_header,
         )
 
     def start_frontend(self, registry: "ProcessRegistry") -> list[ManagedProcess]:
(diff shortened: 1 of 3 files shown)


YAMY1234:
Superseded by #236 — re-done as a stacked PR on top of #233 (configurable header) instead of a standalone PR against main.

PR #240, opened 2026-07-03
touch log after queueing job

nvswarren:
The `apply` sub-command prints a `tail` command that may be used to view the job's log. If the user executes this too quickly, the job may not have started, which will cause `tail` to fail because the file doesn't exist. Fix srtctl to touch the file so that it always exists, so that the `tail` can be run at any time.

### DIFF
--- a/src/srtctl/cli/submit.py
+++ b/src/srtctl/cli/submit.py
@@ -740,9 +740,17 @@ def submit_with_orchestrator(
             metadata=metadata,
         )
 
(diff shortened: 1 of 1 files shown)


nvswarren:
Tested: `tail` command certainly works immediately now. However, the job is taking eons to actually launch, so I haven't yet validated whether (once it does) `tail` shows the actual job content.

EDIT: Of course, as soon as I typed that the job launched, and thus validated that the `tail` command does show the job log as expected.

PR #241, opened 2026-07-04
fix(sa-bench): fix rollup key mapping for p99 latencies, token counts, and ISL/OSL

arpera:
Rollup key mapping is wrong for `sa-bench`. Some of the fields were affected by this (for example, ISL/OSL) resulting in being `None` in the final json statistics file.

### Testing Done

* Manual testing of `sa-bench`  after applying this patch showed no `None` fields in json statistics file.
* `make check` -- no issues found


tests/test_sa_bench_rollup.py
```diff
@@ -32,47 +32,43 @@ def test_sa_bench_rollup_generates_json_and_csv_without_metadata(tmp_path):
 
     valid_low_concurrency = {
         "model_id": "GLM-5-FP8",
-        "random_input_len": 8192,
```
KaunilD, line 35: hold up. they've removed isl and osl from the spec ? this a load bearing metric for fixed isl osl workloads. the conforms  with agentic workloads where total tokens matter more. let me confirm from the latest ci run from inferencex.

KaunilD: yeah this as been moved from random_input_len -> isl
https://github.com/SemiAnalysisAI/InferenceX/actions/runs/28916955742/artifacts/8164384297

src/srtctl/benchmarks/scripts/sa-bench/rollup.py
```diff
@@ -35,14 +35,15 @@
...
-    return None
+_ISL_OSL_RE = re.compile(r"sa-bench_isl_(\d+)_osl_(\d+)")
+
+
+def _parse_isl_osl(dir_name: str) -> tuple[int | None, int | None]:
+    """Parse ISL/OSL from the result dir name; (None, None) if absent."""
+    m = _ISL_OSL_RE.search(dir_name)
+    if not m:
+        return None, None
+    return int(m.group(1)), int(m.group(2))
```
KaunilD: @arpera  this directory name patter could be volatie. i think its better to check the json contract instead of the regexp.

tests/test_sa_bench_rollup.py
arpera: Hi Kaunil, probably I misunderstand something but I did this PR only targeting sa-bench implementation from srt-slurm (not InferenceX's sa-bench). I suppose InferenceX has it's own implementation of sa-bench and this implementation differs from srt-slurm's one (particularly InferenceX's one has isl/osl field in results json file, but srt-slurm's one does not have them). If I understand right then srt-slurm's implementation of sa-bench does not have isl filed in json file. So, my point is that in this PR I am attempt to fix srt-slurm's sa-bench results parsing not InferenceX's sa-bench. Could you tell me please what I should do then?

src/srtctl/benchmarks/scripts/sa-bench/rollup.py
```diff
@@ -220,26 +221,27 @@ def main(log_dir: Path) -> None:
             continue
 
         if not config:
+            isl, osl = _parse_isl_osl(result_file.parent.name)
```
KaunilD: alright. i have enough context. sa bench decided to not report isl osl. as suspected. in that case @arpera  lets get the isl osl from metadata.benchmark.{isl, osl}. also lets add a test such that for agentic workloads ONLY this isl osl fields are none. for fixed workloads sa-bench it should be integer. attached example below.

```json
  "benchmark": {
    "type": "sa-bench",
    "isl": 8192,
    "osl": 1024
  },
```

src/srtctl/benchmarks/scripts/sa-bench/rollup.py
```diff
@@ -220,26 +221,27 @@ def main(log_dir: Path) -> None:
...
+            "tpot_p99_ms": data.get("p99_tpot_ms"),
             "itl_mean_ms": data.get("mean_itl_ms"),
-            "itl_p99_ms": _get_percentile(data.get("percentiles_itl_ms", []), 99.0),
+            "itl_p99_ms": data.get("p99_itl_ms"),
             "e2el_mean_ms": data.get("mean_e2el_ms"),
             "completed_requests": data.get("completed"),
-            "total_input_tokens": data.get("total_input"),
-            "total_output_tokens": data.get("total_output"),
+            "total_input_tokens": data.get("total_input_tokens"),
+            "total_output_tokens": data.get("total_output_tokens"),
```
KaunilD: same for metric parsering as well. a new version of srt slurm can be used with older version of sa bench. in that case lets do a safe get like safe_get(data, ["alias_1", "alias_2"], default_if_None) if alias_1 is presen it will return alias 1 else alias_2.

src/srtctl/benchmarks/scripts/sa-bench/rollup.py
KaunilD: this contract is supplied by the user so its going tp be more robust and sustainable.

KaunilD (changes requested):

src/srtctl/benchmarks/scripts/sa-bench/rollup.py
arpera: Fixed

src/srtctl/benchmarks/scripts/sa-bench/rollup.py
arpera: Fixed

src/srtctl/benchmarks/scripts/sa-bench/rollup.py
arpera: Fixed

vadiklyutiy:
@KaunilD It looks like @arpera fixed you comments. 
Could you pls take a look again?

PR #260, opened 2026-07-12
fix(sa-bench): reuse Dynamo HTTP connections

Thunderbeee:
## Summary

- reuse one benchmark-scoped `aiohttp.ClientSession` for SA-Bench Dynamo requests
- use `TCPConnector(limit=0)` so aiohttp's default connection cap does not override SA-Bench's explicit concurrency
- close the shared pool after the timed request phase and before CPU-heavy metrics processing
- cancel and await pending request tasks before closing the session on failure
- add focused tests for reuse, event-loop ownership, close ordering, failure, and cancellation

## Root cause

The previous Dynamo request function created a new `ClientSession` for every request. Because each session owned its own connector, every request also created and closed its own HTTP connection pool and TCP connection. At high concurrency this caused connection/ephemeral-port churn and bursty admission into the serving frontend.

The client concurrency semaphore remained at 17,920, but part of those closed-loop slots repeatedly went through connection setup and teardown instead of maintaining a stable server-ready backlog. This reduced continuous-batching efficiency: prefill batches were underfilled, decode batches were smaller, and CUDA-graph padding increased.

The fix keeps request payloads and response parsing unchanged while extending only the transport pool lifetime to the benchmark invocation. Warmup and formal phases remain separate Python processes and therefore use separate pools.

## Controlled production validation

All runs used the same frozen model revision, server topology/configuration, 179,200-request token-length trace, 17,920 concurrency, and 330,456,434 total input+output tokens.

| Client | Duration | Total token throughput |
|---|---:|---:|
| Previous built-in SA-Bench | 704.036 s | 469,374.473 tok/s |
| Custom shared-session reference | 553.556 s | 596,970.481 tok/s |
| Fixed built-in SA-Bench | 554.707 s | 595,731.988 tok/s |

The fixed built-in client improved total throughput by 26.92% and finished within 0.21% of the shared-session reference. All 179,200 formal requests succeeded with zero request errors.

Cross-layer evidence:

| Metric | Before | After |
|---|---:|---:|
| Formal unique source-port values | 24,757 | 17,920 |
| Average requests per observed port | 7.24 | 10.00 |
| Prefill token fill | 59.72% | 95.67% |
| Prefill iterations | 4,271 | 2,648 |
| Mean useful decode batch | 290.88 | 437.87 |
| Decode padding overhead | 19.00% | 6.08% |
| Decode iterations | 141,753 | 94,161 |

The fixed connection pattern matched the reference: 179,200 formal requests reused exactly 17,920 source-port values, or 10 requests per connection slot on average.

## Tests

- focused lifecycle suite: `7 passed`
- full repository suite on current `main`: `848 passed, 2 skipped, 6 deselected`
- new-test Ruff check: passed
- Python compilation and `git diff --check`: passed

The tests cover:

- unlimited connector capacity under SA-Bench's own semaphore
- multiple requests reusing the exact injected session
- session closure before metrics processing
- cleanup after ordinary failure and cancellation
- separate sessions for separate event loops
- pending request cancellation before session closure

## Scope

This change does not modify the model, serving configuration, topology, warmup policy, prompt generation, request payload, streaming/SSE parsing, token accounting, throughput formula, or non-Dynamo backends.

## Provenance

The production experiment used commit `946a9be1d100264b7e1532700a35d789c5cae7c3`. This PR replays the identical patch on the current NVIDIA `main` with the required DCO sign-off. Both commits have stable patch-id `35a871ebf85bb421a9480588ed3684112ead4455`.


khubaib92:
Adding some background. We are trying to submit Qwen3.5 TRTLLM fp4 GB300 disagg for SAv3 InferenceX. 
Locally we were running the exps on trtllm-serve with a slightly modified version of benchmark_serving.py client from https://github.com/kedarpotdar-nv/bench_serving (this is a direct fork of official SA benchmark client: https://github.com/SemiAnalysisAI/InferenceX/blob/main/benchmarks/benchmark_lib.sh). 
We were seeing good throughput. 
However, when we ran the same config on InferenceMax using srt-slurm + trtllm-serve, we saw ~30% lower throughput. This PR is the debug/fix to recover the perf. 
The config is 5ctxDEP2 + 7genDEP4 (total 38 gpus), concurrency=17920. 
Note that we saw this perf regression only on high conc pareto points: the mid and low conc points were not affected.

richardhuo-nv:
Let's make this as a flag to be configurable, rather than a default for all submission.

richardhuo-nv (approved):

PR #265, opened 2026-07-15
feat(mooncake): manage standalone store services

00fish0:
## Motivation

In a typical disaggregated SGLang deployment, prefill and decode nodes have different host-memory usage patterns.
Prefill nodes usually enable HiCache. Its L2 cache consumes part of the host DRAM, leaving less memory available for Mooncake Store. Decode nodes normally do not enable HiCache, so more of their remaining allocatable DRAM can be assigned to Mooncake Store.

As a result, the two roles need different Store capacities. For example:

- Prefill: HiCache L2 plus a smaller Mooncake Store segment.
- Decode: no HiCache L2 and a larger Mooncake Store segment.

The existing embedded-client configuration is insufficient for this deployment because the same Mooncake environment is injected into all inference workers. Running standalone Store services allows prefill and decode nodes to use separate MOONCAKE_GLOBAL_SEGMENT_SIZE values.

This also makes memory ownership explicit: the SGLang worker initializes its embedded Mooncake clients without contributing another global DRAM segment, while the standalone Store service owns the role-specific DRAM allocation.

## Summary

This PR extends the existing Mooncake integration for both SGLang and vLLM by:

- Adding managed standalone Mooncake Store services on prefill, decode, or aggregated worker nodes.
- Allowing role-specific Store memory sizes, environment variables, and command-line arguments.
- Adding configurable `mooncake_master` tuning arguments through `mooncake_kv_store.master_args`.
- Showing the resolved Mooncake configuration in dry-run output.
- Documenting the configuration and lifecycle behavior.

## Standalone Store services

A recipe can launch one standalone Store service per selected physical worker node:

```yaml
backend:
  type: sglang

  prefill_environment:
    MOONCAKE_GLOBAL_SEGMENT_SIZE: "0"
  decode_environment:
    MOONCAKE_GLOBAL_SEGMENT_SIZE: "0"

  mooncake_kv_store:
    standalone:
      command: [python, -m, mooncake.mooncake_store_service]
      args: [--port, "8800", --max-wait-time, "120"]
      env:
        MOONCAKE_PROTOCOL: rdma
        MOONCAKE_DEVICE: "mlx5_0,mlx5_1"
        MC_GID_INDEX: "3"
        MOONCAKE_LOCAL_BUFFER_SIZE: "0"
      placements:
        prefill:
          env:
            MOONCAKE_GLOBAL_SEGMENT_SIZE: "100gb"
        decode:
          env:
            MOONCAKE_GLOBAL_SEGMENT_SIZE: "400gb"
      health_check:
        port: 8800
        timeout_seconds: 120
 ```   
   
The services are started after mooncake_master becomes ready and before inference workers are launched.

### CODE REVIEW

**Review by ishandhanani — APPROVED** [review]

**Comment by ishandhanani** [comment]
pls rebase

**Comment by KaunilD** [comment]
Hi please rebase.

**Comment by 00fish0** [comment]
@ishandhanani @KaunilD, apologies for the delayed response — I've now rebased the branch onto the latest main and resolved the conflicts. The PR should be ready for re-review. Let me know if anything else needs adjustment. Thanks for your patience!

File: src/srtctl/cli/do_sweep.py [hunk 1]
```diff
@@ -297,6 +306,134 @@ def start_mooncake_master(self, registry: ProcessRegistry) -> ManagedProcess | N
...
+                logger.info("Starting Mooncake standalone Store service for %s on %s", role, node)
+                logger.info("Mooncake standalone command: %s", shlex.join(command))
+                proc = start_srun_process(
+                    command=command,
+                    nodelist=[node],
+                    output=str(service_log),
+                    container_image=container,
+                    container_mounts=self.runtime.container_mounts,
+                    env_to_set=env_to_set,
+                    bash_preamble=preamble,
```
YAMY1234, line 408: The documented multiline `preamble: |` retains a trailing newline, so passing it through here causes the downstream `" && ".join(...)` to generate a new shell line starting with `&&` and fail before the Store starts. Could we strip or wrap the preamble before composing the command, and add a regression test for the documented YAML block?

YAMY1234:
I validated the PR head (`4c72e831`) on a real 3-node GB300/aarch64 SLURM deployment: one dedicated infra node, one SGLang prefill node, and one decode node, using Mooncake `0.3.12.post1` over RDMA. After omitting the malformed multiline preamble and applying the cluster-specific DMA-BUF setting (`WITH_NVIDIA_PEERMEM=0`), the 100 GiB + 400 GiB standalone Stores registered successfully; a client with `MOONCAKE_GLOBAL_SEGMENT_SIZE=0` completed a byte-exact 1 MiB put/get, and two OpenAI-compatible P/D requests succeeded.

The master observed 58 keys / ~23 MiB allocated across both Store segments, so this exercised the real Store data path rather than only service readiness. The standalone orchestration looks sound for the tested SGLang path; the only PR-code issue I found is the multiline preamble composition noted in the inline review. The DMA-BUF setting and inability to raise the documented `ulimit` values were specific to the tested cluster/container; I did not validate the vLLM path in this run.

src/srtctl/cli/do_sweep.py
00fish0: Fixed by trimming trailing whitespace from the rendered standalone preamble before passing it to start_srun_process, which preserves multiline commands while avoiding a dangling `&&`. Added a regression test using the documented YAML block scalar.

00fish0:
@ishandhanani @KaunilD @YAMY1234  I've already rebased onto the latest main, resolved the conflicts, and fixed the multiline preamble issue that @YAMY1234 caught (added a regression test for it). The standalone Mooncake Store orchestration should be ready for final review. Thanks!

ishandhanani:
so sorry - can you rebase again. after that ill merge. can you ping me in mooncake slack please

ishandhanani:
Closing with the 2.0 merge (#407, part 11): standalone Mooncake stores are `type: mooncake-store` services (one entry per role), and the master is the implied `mooncake-master` service. See docs/services.md. Thank you for the design; it shaped the typed-service model.

PR #269, opened 2026-07-16
fix(vllm): increase DeepGEMM NVLink barrier timeout for GB200

Ankur-singh:
## Summary

- Increase the vendored DeepGEMM NVLink barrier timeout from 30 seconds to 300 seconds for GB200 vLLM startup.
- Add a vLLM setup wrapper that runs the existing container dependency setup before applying the timeout patch.
- Make the DeepGEMM patcher idempotent and fail safely when the vendored source has drifted or contains ambiguous anchors.
- Prevent multi-node vLLM workers from inheriting a shared `VLLM_PORT`, allowing vLLM to choose an ephemeral base port for its internal message queues.
- Add generic `env_to_unset` support to `start_srun_process()`, applied after explicit exports and before the setup preamble and worker command.
- Add tests for timeout patching, idempotency, source-drift handling, wrapper ordering, environment-unset ordering, and multi-node vLLM behavior.

## Motivation

This addresses two startup failure modes seen with multi-node GB200 vLLM deployments:

1. DeepGEMM initialization can exceed its default 30-second NVLink barrier timeout. The setup patch raises that timeout to 300 seconds before vLLM imports the vendored DeepGEMM sources.
2. Tensor-parallel workers spanning multiple nodes can inherit the same `VLLM_PORT`. Because vLLM uses this value as the base port for internal message queues, concurrent workers can race while probing and binding ports. The worker stage now unsets `VLLM_PORT` only for vLLM endpoints spanning more than one node so vLLM can select an available ephemeral port.

Recipe selection remains in downstream repositories and is intentionally outside this PR.

## Provenance

- DeepGEMM timeout fix ported from @jthomson04's [original source branch](https://github.com/jthomson04/srt-slurm/tree/jthomson04/vllm-gb200-v0.1.dev18219%2Bg4ba0a72a4) and [source commit `0835fd5`](https://github.com/jthomson04/srt-slurm/commit/0835fd5141f6bf5b5e3b6d53850a84617bff23a5).
- The follow-up multi-node `VLLM_PORT` fix was also authored by @jthomson04 in [commit `9908c4d`](https://github.com/NVIDIA/srt-slurm/commit/9908c4ddffa3d6ffc1ff36a740d4d8f9610037a4).

## Validation

- `uv run pytest tests/test_vllm_deepgemm_timeout_fix.py tests/test_slurm.py -q` — 23 passed
- `uv run ruff check configs/patches/vllm_deepgemm_timeout_fix.py src/srtctl/cli/mixins/worker_stage.py src/srtctl/core/slurm.py tests/test_vllm_deepgemm_timeout_fix.py tests/test_slurm.py`
- `bash -n configs/patches/vllm-container-deps-deepgemm-timeout.sh`
- `git diff --check origin/main...HEAD`
- GitHub CI at head `5993f75`: `lint`, `typecheck`, `test`, `mock-and-server`, `validate-recipes`, and `copyright-check` passed.

### DIFF
--- a/configs/patches/vllm-container-deps-deepgemm-timeout.sh
+++ b/configs/patches/vllm-container-deps-deepgemm-timeout.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+set -euo pipefail
+
+bash /configs/patches/vllm-container-deps.sh
+python3 /configs/patches/vllm_deepgemm_timeout_fix.py
--- a/configs/patches/vllm_deepgemm_timeout_fix.py
+++ b/configs/patches/vllm_deepgemm_timeout_fix.py
@@ -88,4 +88,3 @@ def main() -> None:
 
 if __name__ == "__main__":
     main()
-
--- a/configs/patches/vllm_deepgemm_timeout_fix.py
+++ b/configs/patches/vllm_deepgemm_timeout_fix.py
@@ -1,0 +1,91 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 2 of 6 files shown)


Ankur-singh:
Hi @jthomson04 — could you please review this clean port of your DeepGEMM timeout fix? The PR description links your original branch and source commit.

PR #270, opened 2026-07-17
Add fields for working GPU count in rollup data and use working GPU count to compute total_throughput_per_gpu

HuiGao-NV:
We may allocate part of devices on one node, such as 1 prefill server with DEP4 only needs 4 GPUs.
With current formula, this server will be counted with 8 GPUs although it only use 4.
The total throughput per gpu is more accurate to be computed with the real number of working GPUs, so add new logic to get the number of real working GPU devices.

### DIFF
--- a/src/srtctl/benchmarks/scripts/sa-bench/rollup.py
+++ b/src/srtctl/benchmarks/scripts/sa-bench/rollup.py
@@ -21,6 +21,9 @@ OUTPUT_FIELDS = [
     "Config",
     "Total GPU Count",
     "Decode GPU Count",
+    "Total Working GPU Count",
+    "Decode Working GPU Count",
+    "Prefill Working GPU Count",
     "Concurrency",
     "Total Token Throughput",
     "Output Token Throughput",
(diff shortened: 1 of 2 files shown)


HuiGao-NV:
[669968.tar.gz](https://github.com/user-attachments/files/30332485/669968.tar.gz)
This is the raw data of one slurm job.
The changed data fields are in 669968/logs/benchmark-rollup.csv:

> Config,Total GPU Count,Decode GPU Count,_Total Working GPU Count,Decode Working GPU Count,Prefill Working GPU Count_,Concurrency,Total Token Throughput,Output Token Throughput,Median TTFT,Median TPOT,Median ITL,P90 Decode Running Requests,Output Token Throughput per User,Total Token Throughput per GPU
> disagg-B300-1p5d-dep4-tep4-c195-stp,32,24,24,20,4,195,86848.766,9631.879,3286.376,15.534,1548.842,,64.377,3618.699

We can see that the "Total GPU Count" is 32 while the "Total Working GPU Count" is 24 with 4 prefill GPUs sand 20 decode GPUs. This is consistent with config in 669968/config.yaml:

> resources:
>   gpu_type: B300
>   gpus_per_node: 8
>   prefill_nodes: 1
>   prefill_workers: 1
>   gpus_per_prefill: 4
>   decode_nodes: 3
>   decode_workers: 5
>   gpus_per_decode: 4

PR #273, opened 2026-07-18
fix(benchmark): count per-node vLLM health entries

ivanium:
## Summary

- count Dynamo vLLM health registrations from the physical process topology for per-node DP launch
- preserve existing per-rank counting for per-GPU DP launch
- cover disaggregated and aggregated per-node layouts

## Why

The health gate multiplied logical workers by the global DP size for every vLLM DP layout. In `per_node` launch mode, Dynamo registers one generate instance per node-local process instead, so the gate waited for registrations that would never appear.

## Impact

Per-node vLLM jobs can pass the startup health gate once all actual generate processes are registered. Other backends, frontends, and per-GPU vLLM launch behavior are unchanged.

## Validation

- `make check` (867 passed, 2 skipped, 6 deselected)
- focused health-expectation tests
- `git diff --check`

### DIFF
--- a/src/srtctl/cli/mixins/benchmark_stage.py
+++ b/src/srtctl/cli/mixins/benchmark_stage.py
@@ -46,12 +46,32 @@ def _vllm_data_parallel_size(config: "SrtConfig", mode: str) -> int:
     return int(mode_config.get("data-parallel-size") or mode_config.get("data_parallel_size") or 1)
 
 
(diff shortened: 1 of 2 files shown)


alec-flowers:
Took these changes and added a bit more. Thanks!
[[codex/pr273-force-hybrid-lb](https://github.com/alec-flowers/srt-slurm/tree/codex/pr273-force-hybrid-lb)](https://github.com/NVIDIA/srt-slurm/pull/277)

PR #284, opened 2026-08-01
Add recipes for disagg vllm serving of Qwen3.5 on GB200 8k1k

arpera:
Add recipes for Qwen3.5 disaggregated serving using vLLM on GB200 ISL/OSL = 8k/1k. Uses topologies NxDEP2+1xDEP8, where N is number of prefill endpoints in range 4 to 8. random range ratio set 0.8 as SA submissions require.
These recipes were also mirrored in srt-slurm-recipes https://github.com/NVIDIA/srt-slurm-recipes/pull/3


arpera:
Close this PR because now all the recipes are stored in srt-slurm-recipes

PR #289, opened 2026-08-05
[2/4] feat(telemetry): collect multinode DCGM power into strict artifacts

edwingao28:
> **Stacked series.** This PR's diff includes #287 and #288 until they merge (GitHub cross-fork PRs cannot base on another PR's branch); it will shrink automatically as predecessors land. This rung's own change: https://github.com/edwingao28/srt-slurm/compare/ladder/a1a-power-artifacts...ladder/a1b-power-collector

## What

The collector runtime for multinode DCGM power telemetry: a per-node exporter scrape loop that writes the artifact set introduced in #288 (samples CSV, expected windows, `manifest.json`) during a sweep.

## How it works

- `core/power/session.py` — `PowerTelemetrySession`: one scrape cycle per configured frequency across all node endpoints; per-cycle deadline-abandonable daemon threads (a hung endpoint must not stall the cycle; `ThreadPoolExecutor` workers are non-daemon and joined at exit, which is exactly the hang this avoids); strict accounting — every miss lands in the manifest as a reason code instead of being dropped, including a poller still alive at the cycle deadline (abandoned workers are counted as `endpoint_timeout`, not silently skipped).
- `cli/mixins/telemetry_stage.py` — exporter container lifecycle (srun), readiness probe, session start/stop around the benchmark stage, producer commit resolution via `git_state` (with a wrong-repo guard for site-packages installs nested in a foreign git tree).
- `cli/do_sweep.py` — dispatches telemetry providers by config value; `EVAL_ONLY=true` skips the provider entirely (eval-only runs skip the benchmark, so every expected window would be missing and required-mode telemetry would fail an otherwise successful evaluation).
- `core/schema.py` — `dcgm-power` provider config; rejects `default_frequency` above the sample-gap limit (a frequency above `MAX_SAMPLE_GAP_SECONDS` would fail every window by construction).
- `core/validation.py` — preflight classifies bare registry URIs (`nvcr.io/...:tag`) the same way the existing image classifier does, instead of only `://` forms.
- Multinode exporter srun sizing: `--nodes` now matches the nodelist length (the previous single-node form with a multinode nodelist was rejected by Slurm); pinned by a regression test.

## Failure semantics

Best-effort mode records reasons and never fails the job; `required` mode fails the job when publication is invalid. Operational failures (exporter never ready, scrape loop crash) fail the job in either mode.

## Evidence

This collector has been running in production CI on real multinode deployments:

- GB200 and GB300 1P1D disaggregated deployments (4 prefill + 4 decode GPUs), N=3 independent full concurrency ladders (c1–c128) per platform; **8/8 measurement windows strictly valid in every run**, max in-window sample gap ≤ 1.03 s at 1 Hz.
- Reproducibility across N=3: J/output-token CV ≤ 1.59% (GB200), ≤ 0.67% (GB300).
- The manifest stamps the producer commit; a canary run from the exact tree this series is cut from recomputed the stored rollup offline from raw artifacts and agreed to float precision.
- The configurable exporter port exists because a shared cluster exposed a host-level exporter already bound on the default port — hit in production, not hypothetical.

## Tests

166 tests across the touched suites: collector lifecycle, readiness gate, failure paths, publication, exporter identity, schema validation, and preflight classification.

`pytest tests/` passes (same 3 pre-existing environment-dependent failures as `main` on my machine: fingerprint CPU probe, apply_mock CPU allocation, mcp_spec explain-field).

## Series

- #287 `fix(processes)` (independent)
- **[2/4] this PR** — collector runtime


FrankD412:
## Code Review

### Confirmed Bugs

**`samples.py` (`_has_non_monotonic_device`)** — Same issue flagged in #288: the non-monotonic check uses `<=` instead of `<`. Equal consecutive timestamps — possible at 1 Hz on a 1-second-resolution clock — satisfy the condition and produce a false `TIMESTAMP_NON_MONOTONIC`, blocking publication on valid data.

**`telemetry_stage.py` (`finalize_power_telemetry`)** — `benchmark_child_reaped` is never set anywhere in production code. `finalize_power_telemetry` reads `getattr(self, 'benchmark_child_reaped', None)`, which always returns `None` because nothing in `benchmark_stage.py` or `do_sweep.py` ever sets this attribute. Consequently `allow_window_mutation=reaped is True` always evaluates `False`, and `convert_running_windows` is dead code in production. On any interrupted or failed benchmark run, window JSON files stay `status='running'` rather than being stamped `interrupted`, so the manifest always shows `MEASUREMENT_WINDOW_INCOMPLETE` for interrupted runs. The `BENCHMARK_CHILD_REAP_TIMEOUT` reason code also never appears in any production artifact despite tests covering that path.

### Plausible Bugs

**`session.py` (`_wait_for_readiness`)** — The readiness poll calls `read_samples(self.samples_path)` on every 20ms iteration — up to 1500 times during a 30-second startup window. Each call opens the file, reads all CSV rows, deduplicates via a set, and checks monotonicity. The file grows with each scrape cycle so later iterations parse progressively more rows. On Lustre or NFS clusters with per-operation latency in the tens of milliseconds, the readiness check wall time can far exceed the configured deadline, blocking the benchmark from starting on time.

**`session.py` (`_resolve_endpoints` / `_run_daemon_workers`)** — Abandoned resolver threads can write to `resolved` after `_run_daemon_workers` returns, producing two failure modes:

`_run_daemon_workers` joins each thread with a remaining-time timeout, but does not stop threads that exceed it — they keep running as daemons with a captured reference to the `resolved` closure dict. After the function returns, the main thread immediately reads `resolved.get(node)` for each node:

- **False miss**: A thread writes `resolved[node] = ip` just *after* the main thread reads that key. The node resolved successfully but is excluded from scraping with `ENDPOINT_RESOLUTION_FAILED`.
- **Silent deadline bypass**: A thread writes `resolved[node] = ip` just *before* the main thread reads it. The node took longer than the deadline but is silently included as if it resolved in time — the timeout is violated for that node.

The failure window is narrow on a healthy cluster where DNS resolves quickly. But the deadline exists specifically to handle degraded DNS or misconfigured nodes — on those clusters, threads are far more likely to still be running when `_run_daemon_workers` returns, so the race fires most often in the exact scenario where the protection is most needed.

The fix is a cancellation flag that threads check before writing:

```python
cancelled = threading.Event()

def resolve(node: str) -> None:
    ip = get_hostname_ip(node, ...)
    if ip and not cancelled.is_set():
        resolved[node] = ip

_run_daemon_workers(...)
cancelled.set()  # threads still running past this point won't write

for node in self._nodes:
    ...
```

FrankD412:
## Code Review (follow-up)

### Confirmed Bugs

**`telemetry_stage.py` (`finalize_power_telemetry`)** — `benchmark_child_reaped` is never set anywhere in production code (carry-over from first pass, still unaddressed). `getattr(self, 'benchmark_child_reaped', None)` always returns `None`, so `allow_window_mutation` is always `False` and `convert_running_windows` is dead code. On any interrupted benchmark, windows stay `status='running'` at finalization and the manifest always reports `MEASUREMENT_WINDOW_INCOMPLETE`.

**`telemetry_stage.py` (`finalize_power_telemetry`)** — The `except Exception` block unconditionally does `return 1`, violating the docstring's guarantee that "telemetry never turns a passing benchmark into a crash". If `stop_and_finalize` raises (e.g. disk full writing `manifest.json`) and `required=False`, the job is failed even though the benchmark passed and telemetry was explicitly non-blocking. Fix: `return exit_code if not self.config.telemetry.required else 1`.

**`schema.py` (dcgm-power validation)** — The schema validates `collector_join_timeout_seconds > request_timeout_seconds`, but `collect_once` blocks for up to `2 * request_timeout_seconds + 1` seconds per cycle. With a schema-valid config like `request_timeout=4.0, join_timeout=5.0`, a slow endpoint causes `thread.join(timeout=5.0)` to expire while the collector is still mid-cycle. `COLLECTOR_JOIN_TIMEOUT` fires — an `OPERATIONAL_FAILURE_REASON` — and the job fails even when the benchmark passed. The validation floor should be `2 * request_timeout_seconds + 1`.

**`windows.py` (`_validate_one`)** — When `_check_coverage` returns any reason code, `gaps = {}` wipes all per-device gap magnitudes that were successfully computed. With 8 GPUs where 7 are within tolerance and 1 exceeds `MAX_SAMPLE_GAP_SECONDS`, the manifest emits `per_device_max_sample_gap_seconds: {}` — operators cannot determine which device exceeded the threshold or by how much without re-reading raw `samples.csv`. The computed gaps are diagnostic data and should be preserved even when there are coverage errors.

### Plausible Bugs

**`windows.py` (`_check_result`)** — A small NTP backwards clock jump (`end_unix < start_unix` by less than 0.5s) can produce a window that passes the clock mismatch tolerance check and gets marked `power_coverage_valid=True` with negative per-device gaps in the manifest. `_check_result` computes `wall = end_unix - start_unix`; if `wall` is negative but `abs(wall - duration) <= tolerance`, no `MEASUREMENT_WINDOW_CLOCK_MISMATCH` is raised. An explicit `end_unix > start_unix` guard would close this.

**`schema.py`** — `_is_safe_relative_subpath` is duplicated between `schema.py` (private) and `contract.py` (public), with a comment noting "equality is pinned by tests". However the test suite (`TestStorageSubdirSafety`) only imports from `contract.py` and does not compare the two implementations. A future change to one without the other produces silent inconsistency between submit-time validation and artifact-read-time path checking.

### From Earlier Analysis

**`parser.py` (`parse_power_scrape`)** — The fix broadened the catch from `except ValueError` to `except Exception` with `# noqa: BLE001`. Based on the actual `prometheus_client` source, the only exceptions that can escape `text_string_to_metric_families` are `ValueError` (stable text format parser, all parse errors) and `KeyError` (OpenMetrics parser, when a native histogram struct is missing required fields: `schema`, `zero_threshold`, `zero_count`, `count`, `sum`). `except (ValueError, KeyError)` is the precise fix — `except Exception` is over-broad and would swallow unexpected bugs in the parser that should propagate.

edwingao28:
close this pr as #290 merged

PR #290, opened 2026-08-05
[3/4] feat(power): add the measurement window contract and stage plumbing

edwingao28:
> **Stacked series.** This PR's diff includes #287, #288, and #289 until they merge; it will shrink automatically as predecessors land. This rung's own change: https://github.com/edwingao28/srt-slurm/compare/ladder/a1b-power-collector...ladder/a2-sa-bench-window

## What

Strict measurement windows: a benchmark child stamps window start/end boundaries so the power collector (#289) can integrate energy over exactly the measured interval, not the deployment lifetime.

## Note on the sa-bench migration

Per the direction to shift from the vendored sa-bench to a custom benchmark pointing at the upstream InferenceX repo (as already done for agentX), **this PR does not touch `benchmark_serving.py` at all**. It ships only the benchmark-agnostic half:

- `measurement_window.py` — a standalone module any benchmark child can adopt to stamp windows (~20 lines of integration).
- `benchmark_stage.py` plumbing — passes the windows directory to whatever benchmark runs (the env is set outside any runner-specific branch), and reaps the child on teardown.

Until a benchmark adopts the stamping, a power-enabled run records `MEASUREMENT_WINDOW` reason codes and is not publishable — the collector and artifacts still work, validity just waits on the adapter. I'm happy to write the stamping integration for the custom benchmark when it lands; the one ask is that it be in place before a lane switches over, so power publication stays valid across the migration.

## How it works

- `benchmarks/scripts/sa-bench/measurement_window.py` — `MeasurementWindow`: atomic window-file lifecycle (running → complete / failed / interrupted), stamped from the benchmark child's own clock (`head_node_unix_clock`).
- `cli/mixins/benchmark_stage.py` — passes the windows directory to the child via env; on teardown/SIGTERM unwind reaps the child with `terminate_and_reap` (#287) and records the outcome — an unreaped child may still be writing, so its window is never converted and the reap timeout lands in the manifest as a reason code.

## Failure semantics

- Child exits cleanly → window `completed` with duration.
- Child killed / interrupted → `interrupted`, reason recorded; publication invalid.
- Child unreapable after SIGKILL → window left untouched (a surviving writer must not be overwritten), `BENCHMARK_CHILD_REAP_TIMEOUT` in the manifest.

## Evidence

This exact `measurement_window.py` module produced every window in the production runs cited in #289 (via a sa-bench integration carried on our development branch, intentionally not upstreamed per the migration note above): GB200/GB300 1P1D, N=3 full concurrency ladders, 8/8 windows valid per run — every window `completed` with a child-clock duration that matches the benchmark log to the second.

## Tests

43 tests: window lifecycle and clock stamping, post-start failure boundaries, boolean-typed field traps, SIGTERM unwind reaping, unreaped-child containment.

`pytest tests/` passes (same 3 pre-existing environment-dependent failures as `main` on my machine).

## Series

- #287 `fix(processes)` (dependency of this rung)
- [2/4] #289 — collector runtime
- **[3/4] this PR** — measurement window contract + stage plumbing


src/srtctl/cli/mixins/benchmark_stage.py
```diff
@@ -319,15 +330,26 @@ def _run_benchmark_script(
...
             while proc.poll() is None:
                 if stop_event.is_set():
                     logger.info("Stop requested, terminating benchmark")
-                    proc.terminate()
                     return 1
                 time.sleep(1)
+            self.benchmark_child_reaped = True
             return proc.returncode or 0
         finally:
+            if proc.poll() is None:
```
FrankD412, line 348: **medium** -- Reaping the local `srun` is treated as proof the remote benchmark task is dead, which re-opens the race `convert_running_windows` forbids.

`terminate_and_reap` returns `True` from both the SIGTERM branch and the SIGKILL branch (`processes.py:35-38`), and `start_srun_process` returns a plain `Popen` with no process group -- so SIGKILLing the srun client does not kill the remote step; slurmd tears it down asynchronously.

Scenario: stop is requested, srun ignores SIGTERM for 15s, we SIGKILL it and set `benchmark_child_reaped = True`. `finalize_power_telemetry` then passes `allow_window_mutation=True` (`telemetry_stage.py:248`), `convert_running_windows` rewrites the in-flight window to `interrupted`, and the still-live benchmark task on the bench node writes `completed` over it moments later. That is exactly the case the `convert_running_windows` docstring says must never happen ("a surviving child could otherwise overwrite the file with `completed`"), and it yields a `completed` boundary that was never audited.

Consider only reporting `True` for the graceful-terminate path here, or having `terminate_and_reap` distinguish killed-from-exited for this caller.

src/srtctl/core/power/contract.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+def atomic_write_json(path: Path, payload: Any) -> None:
+    """Replace ``path`` with serialized JSON and leave no partial file behind."""
+    path.parent.mkdir(parents=True, exist_ok=True)
+    fd, temp_path = tempfile.mkstemp(dir=path.parent, prefix=f".{path.name}.", text=True)
+    try:
+        handle = os.fdopen(fd, "w", encoding="utf-8")
+    except BaseException:
+        with suppress(OSError):
```
FrankD412: **low** -- `os.close(fd)` after a failed `os.fdopen` can close a descriptor another thread already owns.

When `io.open(fd, ...)` fails *after* taking ownership of the fd, CPython closes it as part of its own cleanup. This `os.close(fd)` then targets a number the runtime has already released -- and the `PowerCollector` / `PowerScrape-*` daemon threads are concurrently opening sockets and files, so that number can be reused between the two closes and we would close their descriptor instead. The `suppress(OSError)` hides the resulting EBADF but not the mis-close.

`test_atomic_write_closes_raw_fd_when_fdopen_fails` only covers the case where `fdopen` raises *before* consuming the fd, so it would not catch this. Simplest fix is to drop the explicit close and rely on `fdopen`'s own cleanup.

Same pattern in `src/srtctl/benchmarks/scripts/sa-bench/measurement_window.py` `_atomic_write_json`.

FrankD412 (commented):
Reviewed the three commits on this branch (`e94a8439..5844b489`) for correctness bugs.

Most of the diff looks right: the `_scan` duplicate-reporting fix (no more double-report of the first file when 3+ windows share a key), the `SampleWriter` header-failure handle cleanup, the broadened `parse_power_scrape` exception with the new `endpoint_parse_error` code, the `mark_terminal` re-entry guard, and the `<=` -> `<` monotonicity relaxation (a stuck clock still fails via `measurement_window_not_bracketed`) all check out. Changed files are lint- and format-clean, and the four affected test files pass (218 passed; the 3 `test_sa_bench_http_reuse_flag_*` failures are a pre-existing worktree-path issue with `../lib/profiling.sh`, unrelated to this diff).

Two comments inline: one medium, one low.

I also looked at the unwired `measurement_window.py` writer -- nothing imports it and nothing consumes `SRT_MEASUREMENT_WINDOW_DIR`, including at the tip of the ladder (#291). `docs/power-telemetry.md` in #291 documents that as intentional ("No in-tree benchmark stamps measurement windows yet... The adapter belongs with the benchmark"), so not raising it as a finding.

FrankD412 (commented):
Re-reviewed the 10 new commits (`5844b489..2d454bc6`, linear — no force-push). The hardening pass looks good, and I verified it: **256 passed**, ruff check and format clean on all changed files.

What I checked specifically:

- `_run_daemon_workers` rewritten to return `(values, failures)` through a `queue.Queue`, discarding any outcome with `completed_at >= deadline`. The shared-list-plus-lock mutation-after-return hazard is gone, and `test_late_worker_cannot_mutate_the_returned_snapshot` pins it. The `lambda endpoint: self._poll(endpoint, scrape_seq)` in the comprehension is safe — the per-iteration endpoint is passed as the thread arg, not captured.
- Readiness now rides a `threading.Event` set inside `collect_once` under the writer lock (after `append`+`flush`) instead of re-parsing `samples.csv` in a 20 ms spin loop. Same "one complete scrape" semantics, and `_ready_at_monotonic < deadline` correctly rejects a completion landing after the startup deadline. The `_ready_at_monotonic` write happens-before `.set()`, so the cross-thread read is properly ordered.
- `_check_result` now type-checks result timings with `is_finite_number` before comparing, plus an explicit `wall <= 0` guard.
- `_scan`/`convert_running_windows` wrap directory listing in `try/OSError`, `_scan` requires a `.json` suffix, and `_stays_below` catches `OSError`/`RuntimeError` (symlink loops) rather than only `ValueError`.
- Dropping `if coverage_reasons: gaps = {}` is safe — the only downstream consumer (`validate_artifacts.py` in #291) uses gaps solely for a `max_sample_gap_seconds` summary, and producer/validator run the same function, so the stored-evidence cross-check stays consistent.
- The `telemetry_stage` docstring fix now matches the code (`return 1` on finalizer exception), which it previously contradicted.

## My two earlier comments are still open

- **`benchmark_stage.py:344` (medium)** — `processes.py` isn't touched in this range; `terminate_and_reap` still returns `True` from the SIGKILL branch and `allow_window_mutation=reaped is True` is unchanged. The edit here reworded the comment to "the benchmark child is terminated and reaped before telemetry finalizes," which restates the claim the comment was questioning rather than establishing it. Also, `test_unreapable_child_on_unwind_stays_false` makes `proc.wait` raise `TimeoutExpired` for *both* waits, so the SIGKILL-succeeds path (terminate times out, kill works → `True`) is still untested — and that is the path that enables window mutation.
- **`contract.py:136` (low)** — `with suppress(OSError): os.close(fd)` after a failed `fdopen` is unchanged; the only edit to that file was a new constant.

Happy to be told either is a deliberate accept — just flagging that neither looks addressed yet.

One new low-severity note inline.

src/srtctl/core/schema.py
```diff
@@ -1814,12 +1844,78 @@ def _validate_vllm_nsys_profiler_config_not_set(self):
...
+                f"{_DCGM_POWER_MAX_SAMPLE_GAP_SECONDS}s max sample gap the power validator accepts; "
+                "every window would fail sample_gap_exceeded. Set it to the intended collector "
+                "period (e.g. 1.0)."
+            )
+        worst_case_join_seconds = 2 * (
+            2 * telemetry.request_timeout_seconds + _DCGM_POWER_COLLECT_CYCLE_TIMEOUT_GRACE_SECONDS
+        )
+        if (
+            not _is_finite_positive(telemetry.collector_join_timeout_seconds)
+            or telemetry.collector_join_timeout_seconds <= worst_case_join_seconds
```
FrankD412: **low** — the new `collector_join_timeout_seconds` rule makes the *default* invalid for any `request_timeout_seconds >= 2.5`.

Validation requires `join > 2 * (2 * request_timeout + 1)`, and the default moved 10.0 → 12.0, which only clears the bar at the default `request_timeout_seconds: 2.0`:

| `request_timeout_seconds` | `worst_case = 2*(2*rt+1)` | default join `12.0` |
|---|---|---|
| 2.0 (default) | 10.0 | passes |
| 2.5 | 12.0 | **fails** (guard is `<=`) |
| 3.0 | 14.0 | **fails** |

So a config that bumps `request_timeout_seconds` to 3 for a slow exporter and leaves the join timeout alone now fails to load. Impact is small — no in-tree recipe sets either field, and the error message spells out the formula — but the coupling is invisible from the config file. Deriving the default from `request_timeout_seconds` (or noting the dependency in the field docs) would avoid the trap.

src/srtctl/cli/mixins/benchmark_stage.py
edwingao28: fixed in https://github.com/NVIDIA/srt-slurm/pull/290/commits/7dfdd4f048da0bbfb9e2b86deebef5c7e900a861
- separated local process reaping from measurement window mutation safety
- terminate_and_reap now reports both local proecss was eaped and whether SIGKILL was required
- SIGKILL followed by a successful local wait leaves running windows untouched
- add regressive coverage

src/srtctl/core/power/contract.py
edwingao28: fixed in https://github.com/NVIDIA/srt-slurm/pull/290/commits/7dfdd4f048da0bbfb9e2b86deebef5c7e900a861
- both writer use `cloased=False` now
- the caller closes the raw fd exactly once
- test cover early fdopen failure and fd reuse after wrapper construction
I didn't remove the explicit close because an fdopen failure before wrapper ownership is established would then leak the raw descriptor

src/srtctl/core/schema.py
edwingao28: fixed in https://github.com/NVIDIA/srt-slurm/pull/290/commits/7dfdd4f048da0bbfb9e2b86deebef5c7e900a861
- an omitted join timeout now derived from the request timeout
- the resolved defulat remains 12s at a 2s request timeout and becomes 16s at 3s
- added derived val and boundary test coverage

FrankD412 (approved):

FrankD412:
@edwingao28 -- approved, just need to resolve conflicts.

edwingao28:
@FrankD412 thanks, all stacked prs' conflicts have been resolved

PR #291, opened 2026-08-05
[4/4] feat(power): add the offline artifact validator and telemetry docs

edwingao28:
## What

Offline auditor for the power artifact set, plus the telemetry docs. Closes out the series: everything the collector (#289) and measurement windows (#290) publish can be re-validated after the fact, from the artifacts alone.

## Contents

- `core/power/validate_artifacts.py` — re-derives every manifest claim from the raw artifacts (samples CSV, window files, topology) and compares against the stored rollup: device sets, window coverage, gap limits, lifecycle ordering, and the full exporter identity. Stored-vs-recomputed disagreement is a validation failure. It never integrates energy — consumers do that over the validated window, per the producer contract.
- `cli/validate_power_artifacts.py` + `srtctl-validate-power` console script — `srtctl-validate-power --power-dir <dir> --result-root <root> [--expect-role ROLE=COUNT ...]`; exit 0 iff `publication_valid` holds and recomputation agrees. Malformed or duplicate `--expect-role` values are argparse usage errors, not tracebacks.
- `docs/power-telemetry.md` — the on-disk contract, validity semantics, exit-code behavior, and CLI usage.

## Evidence

The validator has been exercised against real artifact bundles from GB200/GB300 1P1D production runs (N=3 full concurrency ladders per platform):

- `srtctl-validate-power` exit 0 with `--expect-role prefill=4 decode=4`; stored and recomputed rollups agree to float precision on every window.
- It also catches real defects: an early run with a broken window boundary failed offline validation before the numbers were used.

## Tests

85 tests: validator agreement/disagreement on each manifest field, lifecycle ordering, exporter identity shape, CLI usage errors, role expectations, malformed-artifact reasons. The validator fixture drives the real `validate_expected_windows` rather than reimplementing its arithmetic.

`pytest tests/` passes (same 3 pre-existing environment-dependent failures as `main` on my machine).

## Series

- #287 `fix(processes)` (independent)
- [2/4] #289 — collector runtime
- **[4/4] this PR** — offline validator CLI + docs


FrankD412:
@edwingao28 -- do you mind updating this branch?

edwingao28:
> @edwingao28 -- do you mind updating this branch?

updated

src/srtctl/core/power/validate_artifacts.py
```diff
@@ -0,0 +1,581 @@
...
+
+    failures: list[str] = _check_wire_contract(manifest)
+
+    try:
+        expected_devices = _expected_devices(manifest)
+        expected_windows = _expected_windows(manifest)
+    except (AttributeError, KeyError, TypeError, ValueError) as exc:
+        return ArtifactReport(ok=False, failures=(f"manifest malformed: {exc!r}",))
+
+    rows, sample_reasons = read_samples(power_dir / SAMPLES_FILENAME)
```
FrankD412, line 163: **The one column that matters -- `power_w` -- is never re-verified, so the validator certifies tampered or corrupted samples as publishable.**

Everything recomputed from `samples.csv` comes from the *identity and timing* columns: `derive_observed_devices` uses `hostname`/`gpu_index`/`gpu_uuid`/`timestamp_unix`, the window audit uses `sample_times`, and `sample_row_count`/`scrape_count` are counts. `power_w` is only checked for finite and `>= 0` in `_parse_row`. Nothing in `manifest.json` records a digest or any aggregate of the samples file.

Concrete scenario, verified against this PR's own fixture: take a fully valid package and rewrite every `400.0` watt reading to `1.0`. Row count, device set, UUIDs, timestamps, gaps and window coverage are all unchanged, so `validate_power_artifacts` returns `ok=True` and `srtctl-validate-power` exits `0` -- certifying a bundle whose energy integral is off by 400x. Same for a partially-rewritten file from a bad post-processing script.

This undercuts the stated contract in `docs/power-telemetry.md` ("The artifact package is self-describing... a reviewer can check a run without access to the live job"). A `samples_sha256` in the manifest, checked here, would close it cheaply -- the producer already reads the whole file in `_finalize_manifest`.

src/srtctl/core/power/validate_artifacts.py
```diff
@@ -0,0 +1,581 @@
...
+@dataclass(frozen=True)
+class ArtifactReport:
+    """Verdict plus the numbers a reviewer wants to see."""
+
+    ok: bool
+    failures: tuple[str, ...]
+    summary: dict[str, Any] = field(default_factory=dict)
+
+    def render(self) -> str:
+        lines = [f"publication_valid: {self.ok}"]
```
FrankD412: **`render()` reports the canary-shape verdict under the `publication_valid` label, so a wrong `--expect-role` makes a publishable package read as invalid.**

`self.ok` is `not failures`, and `failures` includes the optional `_check_topology` assertions driven by `--expect-role` / `--require-distinct-het-groups`. Those have nothing to do with publication validity -- the docs call them "optionally assert an expected job shape for hardware canaries."

Concrete scenario (run against the fixture in this PR):

```
$ srtctl-validate-power --power-dir .../power --result-root .../logs --expect-role prefill=8
publication_valid: False
  ...
  FAIL expected roles ['prefill'], found ['decode', 'prefill']
  FAIL expected 8 prefill GPUs, found 4
```

The stored and recomputed `publication_valid` are both `true` here; only the operator's expectation was wrong. Printing `publication_valid: False` invites the reader to discard good power data. Suggest splitting the two, e.g. a `verdict:`/`ok:` header line plus a separate `publication_valid:` line carrying the actual manifest verdict.

src/srtctl/core/power/validate_artifacts.py
```diff
@@ -0,0 +1,581 @@
...
+    if not isinstance(manifest, dict):
+        return ArtifactReport(ok=False, failures=(f"manifest malformed: not an object ({type(manifest).__name__})",))
+
+    failures: list[str] = _check_wire_contract(manifest)
+
+    try:
+        expected_devices = _expected_devices(manifest)
+        expected_windows = _expected_windows(manifest)
+    except (AttributeError, KeyError, TypeError, ValueError) as exc:
+        return ArtifactReport(ok=False, failures=(f"manifest malformed: {exc!r}",))
```
FrankD412: **This early return discards the wire-contract failures already computed on line 142, contradicting the documented "every failure is printed".**

`failures = _check_wire_contract(manifest)` runs first, but if `_expected_devices`/`_expected_windows` raise, the report is built from a fresh single-element tuple and everything `_check_wire_contract` found is dropped.

Concrete scenario: a manifest that has both a malformed `expected_devices[0].gpu_index` *and* a bad `dcgm_exporter.port` / non-`complete` `status` / bad `producer` string. The operator sees only `manifest malformed: ValueError(...)`, fixes the device entry, re-runs, and only then discovers the exporter-identity problems -- one round trip per class of defect. `docs/power-telemetry.md` promises "otherwise it is `1` and every failure is printed."

```suggestion
    except (AttributeError, KeyError, TypeError, ValueError) as exc:
        failures.append(f"manifest malformed: {exc!r}")
        return ArtifactReport(ok=False, failures=tuple(failures))
```

src/srtctl/core/power/validate_artifacts.py
edwingao28: Fixed in 4a5a7107. 
- The producer now records a mandatory SHA-256 of the finalized samples.csv, and the offline validator rejects missing, malformed, or mismatched digests. 
- Added regression coverage that rewrites every 400.0 reading to 1.0, producer and startup-failure lifecycle coverage.

src/srtctl/core/power/validate_artifacts.py
edwingao28: Fixed in 792a8336.
- ArtifactReport now carries publication_valid separately from the overall validator ok verdict.
- render() reports both validation_ok and publication_valid.
- Added UT coverage showing that an incorrect --expect-role fails validation while the artifact remains publication-valid.

src/srtctl/core/power/validate_artifacts.py
edwingao28: Fixed in 792a8336.
- Preserved wire contract failures on malformed topology.
- Added combined failure coverage.

edwingao28:
@FrankD412 Addressed all comments, could you PTAL?

FrankD412:
Re-reviewed at `792a8336`. All three inline findings are fixed, and I verified each against my original repro rather than by reading the diff:

- **`samples_sha256`** — rewriting every `400.0` reading to `1.0` now fails with `samples_sha256 mismatch: manifest records c2bd3f16..., recomputed fd91ae3a...`. The digest is computed after the writer is closed and `_mutation_disabled` is set, so it covers the final bytes. Making it mandatory in `_check_wire_contract` (rather than "verify if present") is the right call.
- **Verdict split** — a wrong `--expect-role` now prints `validation_ok: False` / `publication_valid: True`, so an operator can no longer misread their own bad expectation as unpublishable data.
- **Preserved wire failures** — a manifest with both a bad port and a malformed `expected_devices` entry now reports both in one run.

Suite is green apart from the 5 pre-existing environment-dependent failures (`sched_getaffinity`, the macOS bash-3.2 `BASH_SOURCE` cases in `test_benchmarks.py`, `test_apply_mock`); `ty` and `ruff` report nothing on any `core/power/` file.

---

One thing I never raised in the original pass, worth considering before merge.

**The publication rule is implemented twice, and no test crosses the boundary between the two implementations.**

`session.py:479` computes it:

```python
publication_valid = (
    status == STATUS_COMPLETE
    and devices.valid
    and windows_valid
    and not sample_reasons
    and not self._manifest.artifact_errors
)
```

`validate_artifacts.py:439` recomputes it, in a deliberately different shape — `status == STATUS_COMPLETE` is re-derived inline as `lifecycle_complete` from the stored `reason_codes` plus `required`, and there is an extra `startup_recovered` gate with no producer counterpart.

To be clear: **these cannot diverge today.** I traced `EXPORTER_LAUNCH_FAILED` and `ENDPOINT_RESOLUTION_FAILED` and both necessarily produce `EXPECTED_DEVICE_MISSING`, so `startup_recovered` is unreachable-but-defensive. That's correct as written. The concern is that the invariant holding it together lives two modules away, in `telemetry_stage.py`, where `nodes` and `expected_devices` are derived from the same `backend_processes` list.

The gap is structural: `validate_power_artifacts` is only ever called from `tests/test_power_validator.py`, and that file never calls `stop_and_finalize`. Its `package` fixture hand-builds a `PowerManifest` and takes `publication_valid` as an input parameter. So the producer's formula and the validator's formula are never run against each other.

The fixture already solves exactly this problem for the window audit — it calls the real `validate_expected_windows`, with the comment *"The audit the producer stores is the one the validator recomputes, so it can never drift."* The publication verdict is the one claim that didn't get that treatment.

The `samples_sha256` fix is a good illustration of the shape of the risk: it added a fourth thing the producer must write for the validator to accept a package. You did add `test_terminal_manifest_records_the_samples_digest` to cover it, which closes that instance — but it closes it because you remembered. A round-trip test closes the next one automatically.

**Suggested fix** — three lines on the end of the existing `test_covered_window_publishes` (`tests/test_power_collector.py:605`), which already builds a complete publishable package:

```python
report = validate_power_artifacts(
    power_dir=session.power_dir,
    result_root=session.power_dir.parent,
)
assert report.ok is True, report.failures
```

I confirmed it passes as-is, and that it has teeth: stubbing out the producer's `samples_sha256` write makes it fail with `assert False is True`. `result_root` is `power_dir.parent` because `_settings` leaves `log_dir` unset, which is where `_write_window_and_result` already puts the result directory.

Asserting `ok is True` is a strong assertion rather than a weak one — `_check_stored_evidence` only returns clean when the stored verdict is `True` *and* agrees with recomputation, so one call pins both formulas together permanently.

Non-blocking; entirely your call whether it's worth it here or as a follow-up.

FrankD412:
@edwingao28 -- Sorry, I was out of the office on Friday. I posted a review last night for you. Otherwise, let me know about that finding and if it's a non-issue we can merge this.

edwingao28:
> @edwingao28 -- Sorry, I was out of the office on Friday. I posted a review last night for you. Otherwise, let me know about that finding and if it's a non-issue we can merge this.

Thanks @FrankD412 . I think you are right on the review, I have added the round trip assertion as proposed. Also updated branch with all tests passed. PTAL when you are free!

src/srtctl/core/power/validate_artifacts.py
edwingao28: resolved

src/srtctl/core/power/validate_artifacts.py
edwingao28: resolved

FrankD412:
## Code Review

### Confirmed Bug

**`session.py` (`_finalize_manifest`)** — A missing `samples.csv` produces both `SAMPLES_CSV_MISSING` and `SAMPLES_CSV_MALFORMED` in the manifest's `reason_codes`. When `samples.csv` never exists (e.g. exporter startup failure, endpoint unreachable), `read_samples` returns `SAMPLES_CSV_MISSING` via its early `is_file()` guard. Then `sha256_file` raises `FileNotFoundError` (an `OSError` subclass), and the `except OSError` block appends `SAMPLES_CSV_MALFORMED` as well.

The offline validator recomputes `read_samples` and gets only `SAMPLES_CSV_MISSING`, so `_check_stored_evidence` sees `stored-only=['samples_csv_malformed']` and reports a spurious "disk-derived reason_codes mismatch" on top of the legitimate failures. Any production run where no samples are collected produces a manifest that fails offline revalidation with a false error.

Fix: guard with `if self.samples_path.is_file():` before calling `sha256_file`, or simply don't catch `FileNotFoundError` — `_check_wire_contract` already rejects `samples_sha256=None` with a format error:

```python
if self.samples_path.is_file():
    try:
        self._manifest.samples_sha256 = sha256_file(self.samples_path)
    except OSError:
        sample_reasons = dedupe([*sample_reasons, Reason.SAMPLES_CSV_MALFORMED])
```

### Plausible Bug

**`validate_artifacts.py` (`_check_topology`)** — The early return on `role_conflicts` silently drops all `--expect-role` count assertions:

```python
roles, role_conflicts = resolve_roles(expected_devices)
if role_conflicts:
    return [f"{reason} (topology)" for reason in role_conflicts]  # count checks never run
```

When the topology has `CONFLICTING_WORKER_ROLES`, the per-role count loop is never reached. A canary invoked with `--expect-role prefill=4 --expect-role decode=4` against a manifest with conflicting roles reports only `conflicting_worker_roles (topology)` — the count assertions are silently skipped. For automated canary alerting that keys on specific failure strings, this means role-count checks are unreliable whenever the topology is also broken.

edwingao28:
Thanks for the review! All fixed in 7f9829f0 @FrankD412 
- A missing samples.csv now remains SAMPLES_CSV_MISSING instead of also being classified as malformed.
- role counts are not well defined after CONFLICTING_WORKER_ROLES, resolve_roles() intentionally returns no role mapping so I don't think emitting syntheic found 0 could count failures. I've add an explicit diagnostic.

FrankD412:
## Code Review (follow-up)

### Remaining Gap in the Missing-Samples Fix

**`session.py` (`_finalize_manifest`)** — The `FileNotFoundError` fix is correct, but the remaining `except OSError` block still fires on any other I/O error (e.g. `PermissionError`) that occurs *after* `read_samples` has already successfully parsed the file. In that case `SAMPLES_CSV_MALFORMED` is added to a manifest where the samples are actually valid — the offline validator re-reads the file successfully, finds no malformed reason, and reports the same spurious "disk-derived reason_codes mismatch" as the original bug, just via a different trigger. The `FileNotFoundError` path is now covered by the new test; the `PermissionError` path is not.

The fix should also guard against this:

```python
rows, sample_reasons = read_samples(self.samples_path)
if self.samples_path.is_file():
    try:
        self._manifest.samples_sha256 = sha256_file(self.samples_path)
    except OSError:
        sample_reasons = dedupe([*sample_reasons, Reason.SAMPLES_CSV_MALFORMED])
```

This way `SAMPLES_CSV_MALFORMED` is only ever recorded when the file exists but can't be read — which is the only case where the offline validator would also see it.

### Open from Previous Comment

**`validate_artifacts.py` (`_check_topology`)** — Still unaddressed: the early return on `role_conflicts` silently skips all `--expect-role` count assertions and `--require-distinct-het-groups` checks. A canary passing both flags gets only `conflicting_worker_roles (topology)` with no indication of whether the count or het-group checks would have passed.

### New Finding

**`validate_artifacts.py` (`_check_wire_contract`)** — Any artifact produced before this PR (where `samples_sha256` is `null`) fails offline validation with "samples_sha256 is not a lowercase SHA-256 digest" — a format-violation message rather than a missing-field message. Operators cannot distinguish "artifact predates the digest feature" from "digest field is corrupted", and there is no migration path for retained pre-PR artifacts. If backward compatibility with pre-PR artifact bundles is a goal, a `null` `samples_sha256` should produce a distinct message (e.g. "samples_sha256 absent — artifact predates digest collection") rather than a format error.

edwingao28:
Fixed in c54c2bb2:
- Digest failures no longer report malformed CSV.
- Conflicted topology checks now report “not evaluated.”
- Missing digests now have a distinct error message.

FrankD412 (approved):
All findings from prior rounds are addressed. The SHA-256 digest tracking,  reason code,  skip messaging, separated / render, readiness event, and resolver deadline queue are all solid. The structural  divergence between the live and offline formulas is a known open item but practically unreachable given resolution failure implies missing observed devices. Good to merge.

FrankD412 (approved):
All findings from prior rounds are addressed. The SHA-256 digest tracking, SAMPLES_DIGEST_UNAVAILABLE reason code, _check_topology skip messaging, separated validation_ok/publication_valid render, readiness event, and resolver deadline queue are all solid. The structural startup_recovered divergence between the live and offline formulas is a known open item but practically unreachable given resolution failure implies missing observed devices. Good to merge.

FrankD412:
@ishandhanani -- We can go ahead with this now. Thanks @edwingao28 for the patience on the back and forth with this PR!

edwingao28:
Thank you! @FrankD412 

> @ishandhanani -- We can go ahead with this now. Thanks @edwingao28 for the patience on the back and forth with this PR!

FrankD412:
@ishandhanani -- gentle nudge here.

FrankD412:
@edwingao28 -- I should be able to merge this now. Mind giving this one more update?

edwingao28:
> @edwingao28 -- I should be able to merge this now. Mind giving this one more update?

@FrankD412 Thanks for updating the branch! This PR is ready to merge whenever you have a chance.

PR #292, opened 2026-08-06
Added GLM5.2 NVFP4 AgentX recipes for GB300

nvjullin:


recipes/gb300-fp4/glm5.2-agentx.yaml
```diff
@@ -0,0 +1,250 @@
...
+    model:
+      repo: nvidia/GLM-5.2-NVFP4
+      revision: main
+    frameworks:
+      dynamo: 1.2.1
+      sglang: nightly-dev-cu13-20260703-75cdaf43
+  resources:
+    gpu_type: gb300
+    gpus_per_node: 4
+  frontend:
```
richardhuo-nv, line 17: can we use this config instead:
```
frontend:
  type: dynamo
  nginx_session_affinity: true
  enable_multiple_frontends: true
  num_additional_frontends: 4
  env:
    PIP_BREAK_SYSTEM_PACKAGES: '1'
  args:
    router-mode: kv
    router-session-affinity-ttl-secs: '3600'
    active-decode-blocks-threshold: None
    active-prefill-tokens-threshold: None
    active-prefill-tokens-threshold-frac: None
  nginx_session_affinity_header: X-Dynamo-Session-ID
```

nvjullin: I am not familiar with this part of the configs and I copied from an old dsv4 config, can you explain the difference?
One thing I am particularly worried about is the explicit number of additional frontends, which is known to bottleneck under heavy load. Is there a reason why we set 4? The default is 9 I believe.

richardhuo-nv: The way you config the dynamo has been deprecated from the latest version. Sorry, the first dsv4 submission happened during the time dynamo team just started the conv aware routing feature, so a lot of breaking changes gets in in the next few weeks. 

If the default is 9, then we use 9. I just want to make sure the X-Dynamo-Session-ID is being passed as a header, and dynamo conv aware routing only accept the X-Dynamo-Session-ID  right now.

recipes/gb300-fp4/glm5.2-agentx.yaml
```diff
@@ -0,0 +1,242 @@
...
+        mem-fraction-static: 0.8
+        context-length: 1048576
+        enable-hierarchical-cache: true
+        hicache-write-policy: write_back
+        hicache-size: 135
+        hicache-io-backend: direct
+        speculative-algorithm: EAGLE
+        speculative-num-steps: 1
+        speculative-eagle-topk: 1
+        speculative-num-draft-tokens: 2
```
zbpatel, line 123: Hey @nvjullin  should the speculative num steps and num draft tokens match between the prefill and decode servers? 

I see that prefill has (1, 2) and decode has (2, 3). Excuse my ignorance here, not familiar with the conventions.

ishandhanani:
Closing with the 2.0 merge (#407): `recipes/` no longer exists in this repository. Recipes live in the downstream InferenceMAX tree (`benchmarks/multi_node/srt-slurm-recipes/`), and the in-repo corpus is the curated `examples/` matrix. `srtctl migrate -f <recipe> --in-place` rewrites a v1 recipe to the 2.0 layout. Please reopen these downstream if they are still current.

PR #296, opened 2026-08-07
(feat) Support multi-node aggregate jobs with direct vllm frontend

arpera:
Add vllm frontend support for aggregated multi-node vllm jobs. Initial work on adding support vllm frontend for aggregated **single-node** vllm jobs has been done in https://github.com/NVIDIA/srt-slurm/pull/278. Then in srt-slurm [fork](https://github.com/functionstackx/srt-slurm-nv) @functionstackx added support for multi-node case https://github.com/functionstackx/srt-slurm-nv/pull/1 (which is the base for this PR but it has been extended a lot; also improves docs).

This vllm frontend support is required to make reproducible Kimi-K3 multi-node recipes uploaded to SA here https://github.com/SemiAnalysisAI/InferenceX/pull/2391

### DIFF
--- a/src/srtctl/backends/vllm.py
+++ b/src/srtctl/backends/vllm.py
@@ -56,7 +56,6 @@ _VLLM_ORCHESTRATION_FLAGS = frozenset(
         "host",
         "port",
         "master-addr",
-        "master-port",
         "nnodes",
         "node-rank",
     }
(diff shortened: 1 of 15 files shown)


functionstackx:
hi @arpera thanks for upstreaming this & improving upon it!

src/srtctl/backends/vllm.py
```diff
@@ -47,6 +47,79 @@
...
+    for key in list(config.keys()):
+        normalized = normalize_vllm_config_key(key)
+        if normalized in flags:
+            config.pop(key)
+            if normalized not in found:
+                found.append(normalized)
+    return found
+
+
+def pop_vllm_orchestration_flags(config: dict[str, Any]) -> list[str]:
```
alec-flowers: There should be some logging that says if we removed flags, what was removed and what it was replaced by. So its easy for a user to debug. Otherwise its just a silent thing and is hard to debug.

arpera: Good point, fixed.

alec-flowers (approved):

PR #305, opened 2026-08-12
feat(benchmark-manifest): portable manifest schema + validator

tanmayv25:
## What this is

A **schema + validator skeleton** for a portable *benchmark manifest*: one
machine-readable file that fully describes a benchmark run, so a different team
can reproduce the native baseline and build a matched image with **no private
paths, out-of-band files, or tribal knowledge**.

This is the first slice of a larger effort to make TRT-LLM → Dynamo benchmark
handoffs reproducible. It defines the target a handoff must hit; the generator
that emits a manifest from a completed run is a separate, not-yet-built piece
(see "Known gaps").

Draft — opening for early review of the schema shape and the validation policy.

## What it adds (`src/srtctl/benchmark_manifest/`)

| File | Role |
|------|------|
| `models.py` | The manifest schema as Pydantic v2 models (source of truth; also emits JSON Schema). Matches the existing `src/srtctl/contract/` idiom. |
| `validate.py` | Schema conformance + cross-field consistency checks, with a stubbed `--online` layer. |
| `errors.py` | Structured issues with stable `IssueCode`s so CI/tests assert on a code, not free text. |
| `export_schema.py` + `schema/benchmark-manifest.schema.json` | Emit + commit the JSON Schema for non-Python consumers. |
| `README.md` | Layout, usage, the check table, and the known gaps stated openly. |

Plus a reference example (`examples/benchmark-manifest/dsv4-pro-8k1k-b300/manifest.yaml`),
tests (`tests/test_benchmark_manifest.py`), and a `srtctl-validate-manifest`
console entry.

```bash
srtctl-validate-manifest examples/benchmark-manifest/dsv4-pro-8k1k-b300/manifest.yaml
```

## What the validator enforces (static, offline)

Every artifact pinned by digest/SHA (image digest, bare sha256 for wheels /
bundle files / effective-config / SBOM, OCI checkpoint digest), full-length
commits, served == requested model, **silent backend fallback**
(`effective != requested`), no private checkpoint paths (absolute or
mount-relative), baked (not runtime-installed) Dynamo wheel, unique result
`point_id`s, capability declarations, and `manifest_version` compatibility.
Hashes are format-checked only; reachability is left to `--online`.

## Known gaps (deliberately visible, not hidden)

- **Generator** — not built. The manifest is meant to be machine-emitted from a
  run's own artifacts; until then the strong silent-fallback / model-name
  guarantees depend on that generator capturing *effective* values from run
  logs. The static checker only confirms recorded-effective == requested.
- **`--online`** — registry/HF/git reachability + compat-window range check emit
  a single visible `ONLINE_NOT_IMPLEMENTED` warning.
- **Dynamo capabilities** (`worker_without_frontend`, `prefill_only_discovery`)
  are in the schema but default to `available: false` until confirmed.
- **Conversion execution** — the schema pins the conversion *contract*; running
  the conversion is a separate ticket.

## Scope

Owns the portable manifest **schema, validator, reference example, and the
image-conversion I/O contract**. Does **not** own the benchmark-tuned defaults,
the comparison-matrix execution / CI conversion job, or the human submission
playbook — those are tracked separately and use the manifest as their interface.

## Testing

- 41 tests: reference example passes; one negative per failure mode (incl. the
  accept-alternative pin branches, warning branches, exact-error-set isolation,
  and a schema-drift guard asserting the committed JSON matches the models).
- `ruff check`, `ruff format --check`, and `pytest` all green.

The schema shape and the validator's severity/enforcement decisions here already
went through an adversarial code-review pass; happy to walk through the
trade-offs.

### DIFF
--- a/examples/benchmark-manifest/dsv4-pro-8k1k-b300/manifest.yaml
+++ b/examples/benchmark-manifest/dsv4-pro-8k1k-b300/manifest.yaml
@@ -1,0 +1,107 @@
+# Reference benchmark manifest — DeepSeek-V4-Pro, 8k/1k, B300 (disaggregated).
+#
+# This is a WORKED EXAMPLE for the schema + validator skeleton. It is structured
+# to pass `srtctl-validate-manifest`. Digests, commits, and SBOM hashes below are
+# illustrative placeholders in the correct format (sha256:<64 hex>, 40-hex sha);
+# a generator will emit the real values from a completed run. See docs.
+manifest_version: "0.1"
+
+metadata:
+  name: dsv4-pro-8k1k-b300
+  description: DeepSeek-V4-Pro disaggregated throughput frontier, 8192/1024, on B300.
+  owners:
+    trtllm: TODO-trtllm-owner
+    dynamo: tanmayv
+
+artifacts:
+  base_image:
+    ref: nvcr.io/nvidia/tensorrt-llm/release:1.3.0rc23
+    digest: "sha256:1111111111111111111111111111111111111111111111111111111111111111"
+  wheels:
+    - name: ai-dynamo
+      version: "1.3.0.dev2026071601"
+      sha256: "2222222222222222222222222222222222222222222222222222222222222222"
+  sbom:
+    srt_slurm_relpath: examples/benchmark-manifest/dsv4-pro-8k1k-b300/sbom/pip-freeze.txt
+    sha256: "3333333333333333333333333333333333333333333333333333333333333333"
+
+source:
+  repo: https://github.com/NVIDIA/srt-slurm
+  branch: main
+  commit: "abcabcabcabcabcabcabcabcabcabcabcabcabca"
+  bundle:
+    - srt_slurm_relpath: recipes/deepseek-v4/8k1k/b300/disagg-b300-8k-2p1d-dep4-dep8-b32-mtp0.yaml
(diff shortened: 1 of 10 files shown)


examples/benchmark-manifest/dsv4-pro-8k1k-b300/manifest.yaml
```diff
@@ -0,0 +1,107 @@
...
+
+source:
+  repo: https://github.com/NVIDIA/srt-slurm
+  branch: main
+  commit: "abcabcabcabcabcabcabcabcabcabcabcabcabca"
+  bundle:
+    - srt_slurm_relpath: recipes/deepseek-v4/8k1k/b300/disagg-b300-8k-2p1d-dep4-dep8-b32-mtp0.yaml
+      sha256: "4444444444444444444444444444444444444444444444444444444444444444"
+  launcher_patches:
+    - "sed ENROOT_REMAP_ROOT=yes -> no in launcher (required for PMIx)."
```
richardhuo-nv, line 36: how to make sure trtllm team will add all the necessary "patches"? especially those things may not be easily showed through srt-slurm recipe. 

Some example I can think of:
1. numa binding command (numactl/taskset command attached)
2. special frontend and benchmark client placements.
3. some ENV var need to be set dynamically based on different nodes/cluster? eg. EFA patches and configs, UCX_NET_DEVICES dynamic setup

basically those settings that will become a new "feature" of srt-slurm.

PR #308, opened 2026-08-13
feat(observability): single knob for analytics data capture

nv-yna:
## Problem

Getting a benchmark run to emit everything the offline perf tooling needs takes several independent knobs — spread across the backend config, three separate env blocks, and the AIPerf invocation. Miss one and the run *looks* instrumented but silently drops a whole data leg, which you only discover hours later at render time.

Concretely, run `2625935` (DSv4, 14 nodes, 1 h) produced **zero** trace spans, because `DYN_LOGGING_SPAN_EVENTS` was never set. One line, one wasted run.

## What this adds

```yaml
observability:
  enabled: true
```

At config-load time this expands into:

| effect | why it's needed |
|---|---|
| `backend.publish_events_and_metrics: true` | creates the worker/frontend `/metrics` surface at all |
| `enable_iter_perf_stats` + `return_perf_metrics` on every engine config | `enable_iter_perf_stats` is what produces the `trtllm_kv_cache_*` occupancy gauges |
| `DYN_LOGGING_SPAN_EVENTS` / `DYN_LOGGING_JSONL` / `DYN_LOG=debug` on prefill, decode **and** frontend | the only source of per-request `SPAN_CLOSED` trace lines |

and at benchmark time, `AIPERF_EXTRA_ARGS` asking AIPerf for its **per-scrape server-metrics JSONL**. That file is already in the schema the offline tooling reads, but AIPerf's default format selection is `json,csv` only — so it is never written unless requested.

Expansion runs on the raw dict inside `load_config`, so every downstream consumer — worker command builder, engine YAML writer, frontend env — sees it with no extra plumbing.

**Every write is a `setdefault`.** An explicit recipe value always wins, which is what makes this safe to switch on globally. The corollary is that a recipe pinning `publish_events_and_metrics: false` defeats the knob silently, so that case now logs a warning.

## Scope

- The in-job RAW `/metrics` scraper that was originally in this PR is split out into **#309**, which stacks on this branch, so the config knob can be reviewed on its own.
- `_get_analytics_benchmark_env` deliberately does **not** derive `AIPERF_SERVER_METRICS_URLS`. `_get_benchmark_env` on `main` already routes those to both built-in AIPerf runners and `benchmark.type: custom` commands, and the custom path needs its `logical_workers_only=True` view so distributed follower ranks aren't advertised as separate engines. Re-deriving them here would clobber that with the physical-process list.

## Verification

Exercised end-to-end on Lyris jobs **2674508** (30 min) and **2674375** (2 h, DSv4 c3, 7 nodes GB300). Both passed a data-presence checklist in a single run:

- **client** 918 profiling requests, `x_request_id` 918/918
- **metrics** 65 405 scrape lines, 5 distinct endpoints, `trtllm_kv_cache_{used,free,max}_blocks` present
- **traces** 41 972 / 16 707 / 15 426 `SPAN_CLOSED` lines across frontend / prefill / decode

The trace leg had been completely empty in the prior run.

Caveat on reproducing that: those runs were staged against `b93276c2`, before `main` gained its `is_custom` metrics-URL routing, and they also carried local patches to InferenceX's `benchmark_lib.sh` (see below). The knob behaviour itself is covered by the unit tests.

## Tests

10 new tests (`tests/test_observability.py`). `make check` green: **927 passed, 2 skipped**.

The load-bearing one is `test_explicit_recipe_values_win` — the `setdefault` contract is what a stale recipe pinning `publish_events_and_metrics: false` depends on.

## Notes for review

- `observability.enable_otel` / `otel_endpoint` are pre-existing fields on this config object and are untouched.
- `config.observability` is read through `getattr` in the benchmark mixin because several existing tests build the config as a `SimpleNamespace`; an optional analytics knob shouldn't become a hard requirement of the benchmark path.
- **Dependency:** `aiperf_export_level` / `aiperf_server_metrics_formats` reach AIPerf via `AIPERF_EXTRA_ARGS`, which needs a passthrough hook in InferenceX's `benchmark_lib.sh` — every AIPerf flag there is currently hardcoded. That hook exists only as a local patch today and needs upstreaming separately before this knob's AIPerf half takes effect on the `benchmark.type: custom` path. The Dynamo/engine half works regardless.

### DIFF
--- a/src/srtctl/cli/mixins/benchmark_stage.py
+++ b/src/srtctl/cli/mixins/benchmark_stage.py
@@ -502,6 +502,45 @@ class BenchmarkStageMixin:
         urls = list(dict.fromkeys(urls)) if logical_workers_only else sorted(set(urls))
         return {"AIPERF_SERVER_METRICS_URLS": ",".join(urls)}
 
+    def _get_analytics_benchmark_env(self, runner: "BenchmarkRunner") -> dict[str, str]:
+        """Client-side AIPerf analytics flags, active only when observability.enabled.
+
+        Scope is deliberately narrow: ``_get_benchmark_env`` already routes
+        ``AIPERF_SERVER_METRICS_URLS`` to both built-in AIPerf runners and
+        ``benchmark.type: custom`` commands, and the custom path must keep its
+        ``logical_workers_only=True`` view so distributed follower ranks are not
+        advertised as separate engines. This helper must not re-derive those
+        URLs or it would clobber that with the physical-process list.
+
+        What it adds is the request for AIPerf's per-scrape server-metrics
+        JSONL. That file is already in the schema the offline tooling reads,
+        but AIPerf's default format selection is json+csv only, so it is never
+        written unless asked for.
+
+        ``--export-level analytics`` is the single-knob form; the explicit
+        ``--server-metrics-formats`` is passed alongside deliberately, so that
+        an AIPerf build without the analytics level still produces the JSONL
+        rather than silently dropping the whole metrics leg.
+        """
+        # getattr, not attribute access: several tests build the config as a
+        # lightweight SimpleNamespace, and an optional analytics knob must not
+        # become a hard requirement of the benchmark path.
+        observability = getattr(self.config, "observability", None)
+        if observability is None or not getattr(observability, "enabled", False):
+            return {}
+
+        env: dict[str, str] = {}
+        extra: list[str] = []
+        if observability.aiperf_export_level:
+            extra += ["--export-level", observability.aiperf_export_level]
+        if observability.aiperf_server_metrics_formats:
+            extra += ["--server-metrics-formats", *observability.aiperf_server_metrics_formats]
+        if observability.aiperf_slice_duration:
+            extra += ["--slice-duration", str(observability.aiperf_slice_duration)]
+        if extra:
+            env["AIPERF_EXTRA_ARGS"] = " ".join(extra)
+        return env
+
     def _get_benchmark_env(self, runner: "BenchmarkRunner") -> dict[str, str]:
         """Get environment variables for the benchmark script."""
         from srtctl.benchmarks.base import AIPerfBenchmarkRunner
@@ -541,5 +580,9 @@ class BenchmarkStageMixin:
             env.update(self._get_aiperf_server_metrics_env(logical_endpoints, logical_workers_only=True))
         if isinstance(runner, AIPerfBenchmarkRunner) and self.config.benchmark.aiperf_package:
             env["AIPERF_PACKAGE"] = self.config.benchmark.aiperf_package
+
+        # observability.enabled: AIPerf analytics flags (server-metrics JSONL).
+        # The metrics URLs are handled above and must not be re-derived here.
+        env.update(self._get_analytics_benchmark_env(runner))
 
         return env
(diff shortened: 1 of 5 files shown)


nv-yna:
@richardhuo-nv could you take a look? Adding you as a reviewer — I only have read access on this repo so I can't set the reviewer field directly.

#308 is the `observability.enabled` config knob; #309 stacks on it with the in-job RAW `/metrics` scraper.

src/srtctl/core/schema.py
```diff
@@ -1019,16 +1019,54 @@ class ObservabilityConfig:
...
+    and, at benchmark time (see ``BenchmarkStageMixin``):
+
+    * ``AIPERF_SERVER_METRICS_URLS`` for ``benchmark.type: custom`` runs, so
+      AIPerf scrapes the workers and not just its auto-detected frontend.
+
+    Every expansion uses setdefault semantics: an explicit value in the recipe
+    always wins, so ``observability.enabled`` is safe to switch on globally.
+
     Attributes:
+        enabled: Master analytics knob. Default: False.
```
richardhuo-nv, line 1044: Wondering if we should have another flag to enable aiperf specifically.  This "enabled" flag can be used to tune worker and frontend knobs. Cause not all benchmark is using aiperf.

src/srtctl/cli/mixins/benchmark_stage.py
```diff
@@ -559,6 +559,45 @@ def _get_aiperf_server_metrics_env(
...
+        env: dict[str, str] = {}
+        extra: list[str] = []
+        if observability.aiperf_export_level:
+            extra += ["--export-level", observability.aiperf_export_level]
+        if observability.aiperf_server_metrics_formats:
+            extra += ["--server-metrics-formats", *observability.aiperf_server_metrics_formats]
+        if observability.aiperf_slice_duration:
+            extra += ["--slice-duration", str(observability.aiperf_slice_duration)]
+        if extra:
+            env["AIPERF_EXTRA_ARGS"] = " ".join(extra)
```
richardhuo-nv: How does this env work? Does aiperf support this ENV var?

src/srtctl/core/schema.py
nv-yna: Agreed on the principle, and it turned out there was a real bug behind the question rather than just a design preference.

`_get_analytics_benchmark_env` took `runner` but never read it, and it was called unconditionally at the bottom of `_get_benchmark_env` — outside the `isinstance(runner, AIPerfBenchmarkRunner) / elif is_custom` branch ten lines above. So `AIPERF_EXTRA_ARGS` was being emitted for every benchmark type, AIPerf or not. Exactly what you're pointing at.

I did go looking for a config field to gate on before deciding. There isn't a good one: `benchmark.aiperf_package` (`None` by default) and `benchmark.aiperf_args` (`{}` by default) are both optional and falsy on an ordinary AIPerf run, so neither can answer "does this run go through AIPerf". The only real signal is `benchmark.type` → runner class via the registry — only `trace-replay` and `mooncake-router` subclass `AIPerfBenchmarkRunner`, plus `custom` as a special case — which is precisely the `isinstance(...) or is_custom` idiom already used above for `AIPERF_SERVER_METRICS_URLS`.

But rather than gate it, I removed the AIPerf leg entirely. With it gone there's nothing left to gate: `expand_observability` only writes span env, `publish_events_and_metrics` and the engine config. `observability.enabled` now does exactly what you described — tune worker and frontend knobs, and nothing else.

src/srtctl/cli/mixins/benchmark_stage.py
nv-yna: thanks for capturing this. it was used for local bypass, no need to me checked in

richardhuo-nv (approved):

PR #309, opened 2026-08-13
feat(observability): in-job RAW /metrics scraper

nv-yna:
> **Stacked on #308** — this PR's base is that branch, so the diff below is the scraper delta only
> (4 files). Merge #308 first; GitHub will retarget this to `main` automatically.

## Problem

Worker and frontend `/metrics` endpoints are served by the live processes. **When the SLURM job ends they disappear and the data is gone** — there is no after-the-fact scrape. Anything wanting a time series of KV-cache occupancy, router queue depth or frontend event-loop pressure has to capture it *during* the run.

AIPerf does scrape on its own `--slice-duration` cadence, but only for the window it is actively benchmarking, and its aggregation is opinionated. This is the raw, unopinionated complement.

## What it does

A daemon thread started with the benchmark stage, gated on `observability.scrape_metrics` (defaults to `observability.enabled`). It appends each response body **verbatim**:

```json
{"timestamp_ns": 1786194627868177662,
 "endpoint_url": "http://node1:8081/metrics",
 "role": "prefill", "worker_id": "node1",
 "text": "<raw Prometheus exposition text, UNPARSED>"}
```

**Capture emits RAW, the processor parses.** That split is the point: a parser fix never requires re-running the job. `timestamp_ns` is wall-epoch, so it aligns with client records and spans on the same clock.

Stdlib only — it runs on the orchestrator's host python, which already reaches every worker node over the job's network (the same URLs srtctl hands to AIPerf).

## Design notes for review

- **Best-effort by construction.** Every failure path is logged and swallowed; an unreachable endpoint costs one line and a rate-limited warning, never the benchmark.
- **Stops on its own signal as well as the shared `stop_event`.** The latter only fires on *abort*, not normal completion — relying on it alone would leave the scraper polling endpoints that are being torn down. There's a test for exactly this.
- **Worker leaders only.** `backend_processes` holds one entry per physical node for multi-node workers, but only rank zero serves the logical worker's `/metrics`; scraping followers would duplicate rows under a misleading `worker_id`.
- **Drift-free pacing.** The loop waits the *remainder* of the interval, so a slow sweep shortens the gap rather than shifting every subsequent tick later.
- Modelled deliberately on the existing `LiveMetricsSnapshotter` (same `try_start_*` entry-point contract, same swallow-everything posture), so the benchmark mixin stays free of analysis-package internals.

## Verification

Ran on Lyris jobs **2674508** (30 min) and **2674375** (2 h, DSv4 c3, 7 nodes GB300):

```
raw_prometheus.jsonl        7625 lines
roles                       frontend / prefill / decode  all present
worker_id set               6100/6100 worker rows
trtllm_kv_cache_* in text   6100 lines
scrape window               4572 s
```

The captured file also round-trips through the existing downstream parser to the expected schema, with `dynamo_component` / `worker_id` labels injected as usual.

## Tests

6 new tests in `tests/test_metrics_scraper.py` covering the RAW line contract, verbatim capture, graceful degradation on an unreachable endpoint, the no-targets no-op, `scraper_enabled` precedence, and the independent-stop behaviour. `make check` green: **936 passed, 2 skipped**.

### DIFF
--- a/src/srtctl/core/schema.py
+++ b/src/srtctl/core/schema.py
@@ -1053,7 +1053,10 @@ class ObservabilityConfig:
             Required when enable_otel is True.
         scrape_metrics: Run the in-job Prometheus scraper. Defaults to the value
             of ``enabled``; set False to opt out while keeping the rest.
-        scrape_interval_seconds: Seconds between scrape sweeps.
+        scrape_interval_seconds: Seconds between scrape sweeps. Default: 1.0.
+            The floor is 0.5; a sweep slower than the interval simply runs
+            back-to-back rather than queueing (see the drift-free pacing in
+            ``RawMetricsScraper``).
         scrape_output: Filename (under the run's log dir) for the RAW capture.
     """
 
@@ -1062,7 +1065,7 @@ class ObservabilityConfig:
     otel_endpoint: str | None = None
 
     scrape_metrics: bool | None = None
-    scrape_interval_seconds: float = 3.0
+    scrape_interval_seconds: float = 1.0
     scrape_output: str = "raw_prometheus.jsonl"
 
     Schema: ClassVar[type[Schema]] = Schema
--- a/src/srtctl/core/schema.py
+++ b/src/srtctl/core/schema.py
@@ -1032,13 +1032,18 @@ class ObservabilityConfig:
     * ``DYN_LOGGING_SPAN_EVENTS`` / ``DYN_LOGGING_JSONL`` / ``DYN_LOG=debug`` on
       prefill, decode and frontend -- per-request ``SPAN_CLOSED`` trace lines.
 
+    and, at benchmark time (see ``BenchmarkStageMixin``):
+
+    * an in-job Prometheus scraper writing ``raw_prometheus.jsonl`` for the whole
+      benchmark window (see ``scrape_*`` below).
+
     Every expansion uses setdefault semantics: an explicit value in the recipe
     always wins, so ``observability.enabled`` is safe to switch on globally.
 
     Scope is deliberately server-side. The knob configures what the workers and
-    frontend *emit*; it does not reach into the benchmark client. Capturing the
-    ``/metrics`` surface it creates is a separate concern, handled by scraping
-    those endpoints directly rather than by asking a client to re-export them.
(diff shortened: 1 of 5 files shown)


nv-yna:
@richardhuo-nv could you take a look? Adding you as a reviewer — I only have read access on this repo so I can't set the reviewer field directly.

#308 is the `observability.enabled` config knob; #309 stacks on it with the in-job RAW `/metrics` scraper.

src/srtctl/core/schema.py
```diff
@@ -1041,6 +1047,10 @@ class ObservabilityConfig:
     enable_otel: bool = False
     otel_endpoint: str | None = None
 
+    scrape_metrics: bool | None = None
+    scrape_interval_seconds: float = 3.0
+    scrape_output: str = "raw_prometheus.jsonl"
```
richardhuo-nv, line 1069: Do we need different path for different workers/frontends? 

Wondering if we just pass in a log dir for that metrics_scraper.py, and it will automatically create a metrics dir in it and dump metrics for different workers and frontends in different file.

nv-yna: Deliberate, and I should have said why in the docstring — you can't tell from the diff.

The single file isn't just a writer convenience: it's the shape the perf dashboard's ingest layer already reads. Its L1→L2 contract is one `raw_prometheus.jsonl`, one JSON line per `(scrape, endpoint)`, with identity carried *in the record* rather than in the path:

```json
{"timestamp_ns": 1786194627868177662, "endpoint_url": "http://node1:8081/metrics",
 "role": "prefill", "worker_id": "node1", "text": "<raw Prometheus text>"}
```

That repo ships its own `src/capture/scrape_raw_metrics.py` producing exactly this, and its only registered metrics processor (`metrics_prometheus`) parses it into the `server_metrics_export.jsonl` schema the panels consume. This scraper is deliberately a drop-in for that — same five keys, same `frontend|prefill|decode` role vocabulary, same `worker_id` = host / `null` for frontend convention. A directory of per-endpoint files would fork the contract and need a second processor upstream.

richardhuo-nv: I see. Thanks for explaining.

richardhuo-nv (approved):

PR #310, opened 2026-08-14
bench: first class telemetry

ishandhanani:
nice


nv-yna:
consider move rust/ to src/ . for example all source file go to src/tachometer

PR #315, opened 2026-08-18
Size per-node data-parallel ranks by tensor-parallel-size

xinli-sw:
`dp_launch_mode: per_node` assumes one GPU per data-parallel rank, so it rejects any endpoint where a rank spans a tensor-parallel group:

```
ValueError: agg data-parallel-size=2 does not match the endpoint's 16 allocated GPUs
```

That is a valid vLLM topology — 2 nodes x TP8, one DP rank per node, EP across all 16 GPUs. vLLM supports it through the hybrid data-parallel load balancer, which this code path already passes (`--data-parallel-hybrid-lb`, `--data-parallel-size-local`, `--data-parallel-start-rank`).

This sizes ranks by `tensor-parallel-size` instead of assuming 1:

- the check becomes `dp_size * tp_size != endpoint.total_gpus`
- `local_dp_size` divides the node's GPUs by `tp_size`

`tp_size` defaults to 1, so existing DP-per-GPU recipes are unchanged.

Tests: added one case for TP8 x DP2 over two nodes (one process per node, ranks 0 and 1) and one for a mismatched product. Existing `per_node` tests pass unchanged.

### DIFF
--- a/src/srtctl/backends/vllm.py
+++ b/src/srtctl/backends/vllm.py
@@ -572,6 +572,11 @@ class VLLMProtocol:
         config = self.get_config_for_mode(mode)
         return config.get("data-parallel-size") or config.get("data_parallel_size")
 
+    def _get_tp_size(self, mode: WorkerMode) -> int:
+        """Get the tensor-parallel-size for a mode, defaulting to 1."""
+        config = self.get_config_for_mode(mode)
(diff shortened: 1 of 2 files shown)


alec-flowers (approved):
Thanks for this! Didn't know vLLM supported mixed tp dp

PR #320, opened 2026-08-20
feat(analysis): add the component performance dashboard

nv-yna:
## What

Adds an offline tool that turns the artifacts an srt-slurm job **already captures** into a single self-contained HTML page with **Overview / Router / Engine / Frontend** tabs, plus a **Log analysis** tab.

It answers component questions the aggregate benchmark numbers cannot: where TTFT went (admission queue vs. routing vs. prefill vs. KV transfer), whether KV cache or in-flight batch was the ceiling, how evenly the router spread load, and whether the frontend event loop was stalling.

## Why this is small

srt-slurm already emits every input this needs — what was missing was the parse-and-render half. `observability.enabled` produces:

- `<log_dir>/raw_prometheus.jsonl`, via `srtctl.analysis.metrics_scraper`, which was written against this exact RAW contract ("Design rule (mirrors `src/capture/` in the perf-dashboard repo): *capture emits RAW, the processor parses*")
- `SPAN_CLOSED` lines on the worker/frontend logs, via `ANALYTICS_SPAN_ENV`
- on AIPerf-driven benchmarks, a per-request client export under `artifacts/<run>/`

Nothing consumed them. This PR adds the two layers that do.

## Shape

| Layer | Path | Role |
| --- | --- | --- |
| L2 ingest | `src/ingest/` | RAW capture → three fixed intermediate schemas (a *bundle*) |
| L3 render | `src/visualization/` | bundle → one self-contained `.html` |

Both are **stdlib-only** and are **not** part of the `srtctl` wheel. They run from a checkout against a finished log directory:

```bash
python3 -m src.ingest.ingest --run-dir outputs/<job_id>/logs --out /tmp/bundle \
    --worker prefill=dep:4:2 --worker decode=tep:4:4

python3 -m src.visualization.build_dynamo_bench_dash /tmp/bundle /tmp/dash.html \
    --frontend-log outputs/<job_id>/logs/<node>_frontend_0.out
```

**No stage of a benchmark invokes them — the job path is untouched.** Wiring this into `PostProcessStageMixin` behind a config knob is a deliberate follow-up, so this PR carries no runtime risk.

## Which tabs a run can populate

| Benchmark type | Overview | Router / Engine / Frontend | Log analysis |
| --- | :-: | :-: | :-: |
| `mooncake-router`, `router`, `trace-replay` (AIPerf) | yes | yes | yes |
| `sa-bench`, `sglang-bench` | **no** | yes | yes |
| any, `observability.enabled: false` | no | no | yes |

Overview needs the client and trace legs joined on `x_request_id`. `sa-bench` runs vLLM's `benchmark_serving.py`, which writes aggregate `results_concurrency_*.json` only — no per-request records to join. The Log-analysis tab exists for exactly that case: it reconstructs per-request TTFT and a stage breakdown from the frontend log alone, and feeds the *same* renderer as the span-sourced panel.

## Vendoring

From `dynamo-benchmark-perf-dashboard` @ `22f49fea243e43403690b38e70a8d4092dec4cc8`. Only what the component dashboard needs was copied. Left behind: the upstream panel dashboard (`dashboard.py`, `template/`), the framework adapters (`src/common/`, `src/trtllm/`, `src/vllm/`), the capture layer (`src/capture/` — `srtctl.analysis.metrics_scraper` already implements that contract), and two processors unreachable from an srt-slurm run (`agentperf` client, live `tempo` scrape). `get_processor` names the valid options if one of those is requested, so the omission fails loudly rather than as a bare `KeyError`.

Deltas applied on top of upstream, all layout/wiring:

1. `ingest.py` under `src/ingest/`, renderer under `src/visualization/`
2. client default glob `artifacts/*/profile_export.jsonl`, matching where `bench.sh` puts the AIPerf export; span-log default `*.out`, matching `<node>_<mode>_w<i>.out`
3. D3 inlined from the vendored sibling by default (`--d3-cdn` opts out), so the page survives being pulled off a cluster or synced to S3
4. `agentperf` / `tempo` registry entries dropped

Both directories are **excluded from `ruff` and `ty`** so the files stay byte-comparable with upstream — reformatting would rewrite ~100 lines of deliberately dense style and destroy the diff they get re-synced against. This follows the existing exclusion for the vendored vLLM benchmark scripts.

## Testing

`tests/test_component_dashboard.py` (14 tests) pins **the seam**, not the vendored internals — the artifact layout srt-slurm writes must be what the vendored defaults look for, which is what breaks on a re-sync. It builds a miniature `<job>/logs/` directory and drives the real ingest and a real renderer subprocess, asserting: worker-label injection into the KV gauges, the `x_request_id` keying of frontend-log records, tab gating per available source, the same-run enforcement that hard-fails a mismatched `--frontend-log`, and that D3 is actually inlined.

```
1244 passed, 2 skipped, 6 deselected in 30.94s   (1230 before this PR)
uv run ruff check src/srtctl/     All checks passed!
uv run ruff format --check        86 files already formatted
copyright check (CI script, run locally)  OK
```

## For reviewers

- **`src/visualization/d3.v7.min.js` is 280 KB of vendored minified JS.** It buys a genuinely self-contained artifact — the HTML is normally read on a laptop after being pulled off a cluster or synced to S3, where a CDN fetch is not guaranteed. An ISC entry was added to `THIRD_PARTY_LICENSES.md`. If you would rather not carry it, dropping the file and defaulting to `--d3-cdn` is a one-commit revert.
- **The upstream repo has no `LICENSE` file and no SPDX headers.** Both repos are NVIDIA-owned and NVIDIA Apache-2.0 headers were added to every vendored file, but a sign-off from the original author (@karenc) on relicensing into this public repo is worth having before merge.
- **Known limitation, documented:** the renderer looks for `trtllm_config_{prefill,decode}.yaml` for the Engine tab's batch ceilings; srt-slurm dumps `<node>_config.json` instead, so the `--max-batch-*` defaults (128/256) apply and are frequently wrong by orders of magnitude. Pass them explicitly. Teaching the renderer srt-slurm's config format is a follow-up.

Docs: `docs/component-dashboard.md`.


nv-yna:
Refinement to the tab-coverage table in the description, verified after opening (`1dbd572a` documents it in `docs/component-dashboard.md`):

The Overview row for AIPerf benchmarks is correct **at default settings**, but it depends on AIPerf's export level, which is worth calling out because the failure is silent.

`profile_export.jsonl` is only written at export level `records` or `raw` — `RecordExportResultsProcessor` raises `PostProcessorDisabled` otherwise. `records` **is** the default (`OutputDefaults.EXPORT_LEVEL = ExportLevel.RECORDS` in `aiperf/config/artifacts.py`, and `_converter_runtime.py` notes "the CLIConfig default is RECORDS"), so the stock `mooncake-router` / `router` `bench.sh` invocations produce it. But a recipe that passes `--export-level summary` keeps `profile_export_aiperf.json` — which is what `rollup.py` reads, so the benchmark results look completely normal — while dropping the per-request export. The `x_request_id` join key goes with it and the Overview tab simply does not appear.

Also confirms the join itself: `MetricRecordMetadata.x_request_id` is a **required** field, and `transports/base_transports.py` sets the `X-Request-ID` header from it — which is what makes the Dynamo spans carry the matching attribute that `traces_spanlog` keys on. Verified against aiperf 0.11.0, the version this repo pulls in as a dev dependency.

PR #322, opened 2026-08-21
Add phase-aware ruter 3P2D lifecycle recipe

ishandhanani:
## Summary
- package ruter with srt-slurm and add its 3-prefill / 2-decode Qwen3 FP8 lifecycle recipe
- materialize separate prefill cache-aware and decode load-only routing decisions
- retain explicit SRTCTL_TACHOMETER overrides in rendered lifecycle scripts

## Validation
- `cargo test --package ruter`
- `uv run pytest tests/test_lifecycle_render.py -q`
- rendered 3P2D lifecycle with `bash -n`
- live try6767 run in progress


ishandhanani:
Closing: ruter is now the merged Python implementation in #319. The Tachometer override fix will be proposed separately on top of current main.

PR #327, opened 2026-08-21
[feat](vllm) Make per-node DP topology-aware

NVShreyas:
## Summary

Makes `dp_launch_mode: per_node` topology-aware for regular `DP x TP x PP` deployments. srt-slurm derives whether each TP/PP replica is node-local or spans multiple nodes from the configured parallel sizes and GPU allocation.

## Why

The process layout is one vLLM process per allocated node. Whether a TP/PP replica fits locally or requires distributed rendezvous is a property of the topology and should not require a separate user-selected launch mode:

- If `TP x PP` fits in the node-local GPU allocation, one process manages one or more local DP replicas.
- If `TP x PP` exceeds the node-local GPU allocation, each DP replica spans multiple node processes participating in one global vLLM rendezvous.

## Behavior

With `dp_launch_mode: per_node`, srt-slurm now:

- Validates that `DP x TP x PP` equals the endpoint GPU allocation.
- Validates that every TP/PP replica divides evenly within or across nodes.
- Creates one process per allocated node.
- For node-local replicas:
  - Computes `--data-parallel-size-local` as `local_gpus / (TP x PP)`.
  - Computes `--data-parallel-start-rank` from preceding node-local DP ranks.
  - Enables `--data-parallel-hybrid-lb` so every node-local process registers with Dynamo.
- For replicas spanning nodes:
  - Passes one shared `--master-addr`, total `--nnodes`, and global `--node-rank` to all processes.
  - Passes one shared `--data-parallel-address` and RPC port.
  - Keeps only global node rank 0 non-headless; every other process is an engine-only participant.
  - Expects one Dynamo DPLB registration per logical endpoint.
- Derives and owns the local-DP, hybrid-LB, and headless flags instead of accepting conflicting recipe values.

Supported regular layouts include:

- DP4 x TP4 on four-GPU nodes: one local TP4/DP rank per node.
- DP8 x TP8 on four-GPU nodes: two nodes per DP rank.
- DP2 x TP8 on four-GPU nodes: two nodes per DP rank.
- DP4 x TP16 on four-GPU nodes: four nodes per DP rank.
- Equivalent layouts using pipeline parallelism through `TP x PP`.

`per_gpu` remains the legacy default, and standard non-DP TP/PP launching is unchanged.

## Validation

- The PR is a single commit rebased onto the latest `main`.
- 49 focused vLLM topology and Dynamo health tests pass.
- Added explicit coverage for DP4/TP4, DP8/TP8, DP2/TP8, DP4/TP16, local PP, distributed PP, and cross-node DPLB health counts.
- Ruff checks pass for the changed source files.
- Full repository suite: 1,284 passed, 2 skipped, and 1 unrelated existing failure in `test_mcp_spec.py` caused by Python 3.14 rendering `Union[str, NoneType]` instead of `UnionType[str, NoneType]`.
- The distributed global-rendezvous/headless path was previously validated through a complete DP4 x TP4 x EP16 C1-C256 benchmark sweep on four four-GPU GB300 nodes with zero request errors. The node-local selection is covered by command-generation and topology tests in this update.

### DIFF
--- a/docs/config-reference.md
+++ b/docs/config-reference.md
@@ -524,8 +524,8 @@ Each worker leader gets a globally unique port starting at 5550:
 ### vLLM DP launch mode
 
 vLLM data-parallel endpoints use one process per GPU by default. Set
-`dp_launch_mode: per_node` to launch one process per node and let vLLM
-manage the local DP ranks in a shared CUDA namespace:
+`dp_launch_mode: per_node` to launch one process per node. srtslurm derives
+whether each TP/PP replica is node-local or spans multiple nodes:
 
 ```yaml
 backend:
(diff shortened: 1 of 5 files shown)


alec-flowers (approved):
Thx for adding this!

PR #328, opened 2026-08-21
feat: render runs captured without observability, and mark the warmup phase

nv-yna:
> ## 🚧 DO NOT MERGE — for review only
>
> This PR currently shows **CONFLICTING**, and the conflict is **inherited, not introduced here**.
>
> `yna/component-perf-dashboard` already conflicts with `main` *without* this commit. Cause:
> **PR #320 was squash-merged** as `f6f9af49`, so `main` re-added `src/ingest/ingest.py`,
> `traces_spanlog.py`, `build_dynamo_bench_dash.py` and `tests/test_component_dashboard.py`
> as new files rather than inheriting the branch's history — git therefore sees `add/add`
> conflicts on all four.
>
> **This blocks the follow-up PR for `yna/component-perf-dashboard` too, not just this one.**
> The fix is to rebase that branch onto current `main`, dropping the two commits already
> squashed in. Not done here: that branch is still being actively committed to, and
> resolving conflicts inside someone else's in-flight commits is a good way to break them
> silently.

## ⚠️ Read this first — branch composition

This branch is **45 commits ahead of `main`**, but only the last one is new here:

| | |
|---|---|
| **New in this PR** | `23f5dbfc` — the single commit described below |
| Inherited | 44 commits from `yna/component-perf-dashboard`, which are **not yet on `main`** |

PR #320 merged only its first two commits (`1dbd572a`). Everything after that is still
unmerged, and this work depends on it (`iter_log`, `panels.py`, `request_trace`), so it
cannot be based on `main` standalone. If the 44 land first, this PR collapses to one
commit. Review `23f5dbfc` in isolation:

```
git diff bc642e6c..23f5dbfc
```

---

## What

Five changes, all found by pointing the dashboard at run directories it had never seen:
the DYN-3873 tokenizer study, and the c2572 three-way Dynamo-vs-trtllm-serve set. None of
those runs set `observability.enabled`, and between them they broke the renderer in four
separate ways.

### 1. AIPerf's own server-metrics export as a second metrics source

`src/ingest/metrics_aiperf_json.py`, selected with `--metrics aiperf-json`.

A run predating `observability.enabled` has no `raw_prometheus.jsonl`, so it had no metrics
leg at all — which is **every already-landed fix and its own baseline**. AIPerf's export
carries per-second timeslices: the same time series in a different container.

Two transpositions are load-bearing and both fail *silently*:

- **Cumulative vs interval.** A single slice cannot tell you which it is. The mode is
  **detected** per series against the run total in `stats`, not assumed. On the reference
  export **1176 of 1178 series reconcile exactly**; the other two by one slice at a run edge.
- **Name suffixes.** AIPerf keys a family by its base name; every panel uses
  `_total`/`_count`/`_sum`.

Streaming throughout — the export exceeds 2 GB, and the per-timestamp regroup spills into
time-contiguous shards because an external `sort` over ~15 M samples gets OOM-killed on a
shared login node.

### 2. A bundle whose only leg is the iteration log now renders

The run window was built from traces + scrapes only. With neither it collapsed to 1 ns, the
`t > run_dur + 60` guard discarded every bin, and **a run with 175,800 real iterations
produced an empty page** and one INFO line reading `0pts`. The Engine tab was also gated on
the scrape stream alone, though it has two independent producers.

This is the trtllm-serve shape — no Dynamo frontend means no `/metrics` surface and no
spans — so a control run could not be placed beside the Dynamo run it exists to be compared
against.

### 3. Both KV panels fall back to the iteration log

`dynamo_component_gpu_cache_usage_percent` and `trtllm_kv_cache_hit_rate` live on the
**worker** endpoint, which is frequently not scraped. The panels emitted a full-length
series of nulls — an empty axis, which reads as a measured zero rather than absent data.

Both quantities are in the iteration log. The derived reuse reproduces the client's own
Prompt Cache Read % to within ~1 pp on three runs (95.88 / 95.76 / 94.27 % derived vs
95.6–97.2 % reported). When neither source exists the panel now says so in words.

### 4. The in-flight batch panels never rendered — on any srt-slurm run

They filtered workers with `startsWith(role)`, but srt-slurm names a worker
`<node>_<role>_w<N>` — the role is in the **middle**. The filter matched nothing and the
builder returns silently on an empty match, so both panels were **absent rather than
broken**, with no warning.

They now also plot the configured `max_batch_size`: as a reference line when it is on a
comparable scale, and as a utilisation figure when it is not (prefill peaks at 11 against a
ceiling of 256, where a line to scale flattens every series onto the axis).

### 5. The end of the load generator's warmup phase, on every time chart

AgentX warmup is **33 %, 35 % and 38 %** of the three c2572 runs — every percentile above
was reading through it.

The boundary is an **optional field on the panel entity**, so a panel carries its own phase
marker alongside its own series and units. Optional is the design: a harness with no warmup
phase must not acquire an invented boundary, and absent must mean *unknown*, never *t=0*.

Resolved from `benchmark_phase` when the client leg is present, else from `warmup.end_ns`
recorded by a new ingest probe — the export is routinely 700 MB and often not copied, and
the boundary is one integer. A boundary outside the run window is refused with a warning.

**That guard immediately earned itself:** it refused a marker at `t+27,379s` on a 5,810 s
run, which turned out to be the iteration log's local→UTC offset defaulting to `+0h` because
a trtllm-serve run has no metrics stream to anchor against. The client export's own span is
now the fallback anchor and the offset derives correctly as `+7h`. **Any `--metrics none`
bundle was affected.**

Also: a third client-export layout (`agentic/<conc>/profile_export.jsonl`, no
`aiperf_artifacts/` level) which hecate writes and the glob did not match.

## Verification

Five runs across two clusters, end to end:

| run | cluster | what it exercises |
|---|---|---|
| `419901` / `432410` | hecate | AIPerf-JSON adapter; the tokenizer-cache A/B |
| `443491` / `444364` | hecate | adapter + warmup marker (35 % / 33 %) |
| `443490` | hecate | iter-log-only render (trtllm-serve control, 1 tab, 17 panels) |

- **12 new tests**; suite **1363 passed**, 2 skipped
- `ruff check` + `ruff format --check` clean on `src/srtctl/`
- CI's copyright-check script run locally — all `.py` carry NVIDIA SPDX headers
- Every generated page verified through `tools/dashboard_render_check.js` (jsdom), 0 JS errors

## Reviewer notes

- **Vendored dirs stay excluded from ruff/ty**, per the existing exclusion — `src/ingest/`
  and `src/visualization/` keep upstream's dense style so the diff against the re-sync
  source stays readable.
- The **`startsWith(role)` fix (4) is the one to look at hardest** — it changes behaviour on
  every existing srt-slurm dashboard, not just these runs.
- `metrics_aiperf_json.py` is stdlib-only and runs under a bare cluster `python3`, matching
  the other L2 processors.


nv-yna:
Superseded by #330.

#328 was opened off a base that had drifted from `main` and went CONFLICTING. Rather than rewrite its history in place, the 43 commits were replayed onto current `main` (`git rebase --onto origin/main`) on a fresh branch — one conflict, in `postprocess_stage.py`, where #319's `_normalize_ruter()` and this branch's `_build_perf_dashboard()` are both pure additions to the same hook; both kept.

#330 carries the same change, merges clean, and adds live-cluster verification of both the `observability.enabled: true` and `false` paths (which turned up two real bugs the unit tests missed). Review there.

PR #342, opened 2026-08-25
feat: add agentx benchmark type for agentic-coding replay

shljessie:
## Summary

AgentX recipes currently require an InferenceX clone, because the replay client (agentic_srt.sh, benchmark_lib.sh, and the utils/aiperf submodule) lives there and is not vendored in srt-slurm. A native agentx benchmark type is proposed upstream in [NVIDIA/srt-slurm#342](https://github.com/NVIDIA/srt-slurm/pull/342); once it lands we can validate these recipes in CI and update them to type: agentx.

### Example

```yaml
benchmark:
  type: agentx
  model_prefix: "dsv4"
  concurrency: 8
  env:
    KV_OFFLOADING: dram
```

### DIFF
--- a/docs/config-reference.md
+++ b/docs/config-reference.md
@@ -652,6 +652,7 @@ benchmark:
 | `router`          | Router performance with prefix caching         |
 | `mooncake-router` | KV-aware routing with Mooncake trace           |
 | `agentperf`       | AgentPerf trajectory replay (agentperf-client) |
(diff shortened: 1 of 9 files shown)


faradawn:
LGTM. cc @ishandhanani

shljessie:
Still work in progress. Will ping again once ready! Want to do more thorough checking to make sure we don't have any issues running it.

PR #350, opened 2026-08-27
[2/3] feat(observability): tachometer scrapes the complement of client-polled endpoints

nv-yna:
## What

Second of the 3-PR observability-consolidation series. **Stacked on #349** — review the last 4 commits; the first commit here is #349's.

`observability.enabled: true` now collects server metrics with the native Tachometer scraper for the whole run, with no `tachometer:` block required, scraping the **complement** of what the benchmark client polls. The legacy Python RAW scraper no longer follows `enabled` (strictly opt-in via `scrape_metrics: true`), removed entirely in PR 3/3.

## Why

- The RAW scraper + aiperf double-polled every worker endpoint by default; the in-tree warning documents a submission that became irreproducible from exactly that. The complement makes single-polling structural instead of advisory: the excluded set is **derived from the same code that injects `AIPERF_SERVER_METRICS_URLS`**, so if the client's endpoint set ever changes, tachometer's complement moves with it.
- Sweep-mode tachometer has never produced output: srtctl pre-created the storage dir the scraper refuses to start into (every attempted run died at startup, e.g. hecate job 452644). Fixed via the `raw/scrape` nested layout the --bash lifecycle already uses.
- Tachometer Parquet was unjoinable (monotonic-relative timestamps only) and lossy (f32 values corrupt counters above 2^24). Rows now carry epoch `timestamp_ns` (Int64) and Float64 values.
- Nothing ingested the Parquet. Two new L2 processors converge on the frozen schema-2 `server_metrics_export.jsonl`: `metrics_tachometer` (parquet; reconstructs `_bucket`/`_sum`/`_count`, injects `worker_id`/`dynamo_component` like the prometheus path) and `metrics_aiperf_jsonl` (aiperf's per-scrape export, preferred over the opinionated aggregate json). Auto-selection: tachometer → raw_prometheus.jsonl → aiperf-jsonl → aiperf-json.

## Behavior notes

- Tachometer is **best-effort** (`critical=False`): a dead scraper costs the capture, never the benchmark. `validate-setup` still fails fast at submit when `bin/tachometer-scraper` is missing; the default bare binary name now resolves against the checkout's `bin/` before $PATH.
- Backend scrape targets are leaders-only (rank zero serves the logical worker's /metrics); frontend/DCGM/node-exporter endpoints always belong to tachometer.
- Explicit `tachometer.enabled: false` opts out; explicit `true` without `observability.enabled` still raises.
- Host sampler re-gated from the RAW knob onto `observability.enabled`.

## Verification

`make check` green (1477 passed), `cargo test --workspace --locked` green (41 tests incl. new timestamp/precision regression guards), recipe validation green (the qwen3-32b ruter recipe simplifies to just `enabled: true`). New tests: complement exclusion, leaders-only, storage-leaf invariant, tri-state resolution, critical=False, binary resolution, dry-run display, both L2 processors (incl. f64 fidelity and source auto-selection).

### DIFF
--- a/docs/config-reference.md
+++ b/docs/config-reference.md
@@ -1042,37 +1042,30 @@ infra:
 
 ## observability
 
-`observability.enabled` turns on the server metrics and trace surfaces and captures raw Prometheus responses during the benchmark window:
+`observability.enabled` turns on the server metrics and trace surfaces and collects them with the native Tachometer scraper for the whole run:
 
 ```yaml
 observability:
   enabled: true
 ```
 
-The default Python scraper writes `<log_dir>/raw_prometheus.jsonl`. To additionally collect parsed Parquet with the native Tachometer scraper:
+Tachometer scrapes the **complement** of what the benchmark client polls: worker endpoints that appear in `AIPERF_SERVER_METRICS_URLS` are left to the client (a worker endpoint is never double-polled — the extra scrape load has previously made a submission irreproducible), while the frontend, DCGM, and node-exporter endpoints are always Tachometer's. On runs whose benchmark has no aiperf client (sa-bench, lm-eval, serve-only, manual), the complement expands to every endpoint.
 
-```yaml
-observability:
-  enabled: true
-  tachometer:
-    enabled: true
-```
-
-Set `scrape_metrics: false` beside `enabled` to use Tachometer without also writing the raw JSONL capture. Otherwise the Python RAW scraper and Tachometer run together.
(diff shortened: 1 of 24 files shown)


nv-yna:
Correction to the "known pre-existing gap" note in this PR's description: **node-exporter meminfo families ARE collected.** Re-inspection of run 2806252's Parquet shows 58 `memory_*` families (MemTotal/MemAvailable/Dirty/HugePages/…) on every node-exporter endpoint. The earlier "absent" finding was an inspection artifact (an alphabetical `[:6]` slice that cut off before `memory_*`). Full per-class inventory from the verification run: frontend 87 families, backend workers 109, DCGM 19, node-exporter 125 — no gaps in the goal's metric classes.

PR #358, opened 2026-08-27
feat(observability): tachometer on by default for every run

nv-yna:
## What

`observability.tachometer.enabled: null` (the default) now resolves to **ON for every run**, decoupled from `observability.enabled`; the requires-master-knob validation is removed. Explicit `enabled: false` still opts out. `observability.enabled` keeps gating what it always gated: worker metrics *content* (TRT-LLM publish flag + engine stats) and the trace/request-trace/host-sampler legs.

## Why draft

Held until a real production A/B rerun proves the default costs nothing: same recipe, no `observability:` block, current main (tachometer off) vs this branch (tachometer on), comparing client-side TTFT/ITL/throughput. Plan is ready; will attach results before marking ready-for-review.

## Cost argument (to be validated by the run)

1 Hz default (#352), best-effort scraper + sidecars (never critical), complement of the client's URL list (no worker double-polling), and without `observability.enabled` the TRT-LLM worker endpoints serve little/nothing — the capture is dominated by the frontend endpoint plus any recipe-configured exporters.

## Notes

- `validate-setup` already requires `bin/tachometer-scraper` unconditionally, so the default-on flip adds no new setup burden.
- Eval-flow integration tests opt out explicitly (they exercise eval, not telemetry).
- `make check`: 1508 passed; the one failure (`TestRunPostEval::test_eval_only_successful`) reproduces on clean origin/main — pre-existing, unrelated (health-mock path, likely from a recent merge).

### DIFF
--- a/docs/config-reference.md
+++ b/docs/config-reference.md
@@ -1112,7 +1112,7 @@ infra:
 
 ## observability
 
-`observability.enabled` turns on the server metrics and trace surfaces and collects them with the native Tachometer scraper for the whole run:
+Tachometer collection is **on by default for every run** (no configuration needed; `observability.tachometer.enabled: false` opts out). `observability.enabled` turns on the server metrics *content* (the TRT-LLM publish flag and engine statistics) and the trace surfaces:
 
 ```yaml
 observability:
@@ -1155,7 +1155,7 @@ observability:
 
 | Tachometer field | Type | Default | Description |
 | ---------------- | ---- | ------- | ----------- |
-| `enabled` | bool/null | `null` | `null` follows `observability.enabled`; explicit `false` opts out; explicit `true` without `observability.enabled` is a validation error |
+| `enabled` | bool/null | `null` | `null` means ON for every run (decoupled from `observability.enabled`); explicit `false` opts out |
 | `binary_path` | string | `tachometer-scraper` | Scraper command or path on the compute nodes |
 | `default_frequency` | float | `1.0` | Scrape frequency in Hz |
 | `sync_interval_secs` | int | `120` | Interval for intermediate Parquet compaction; `0` disables it |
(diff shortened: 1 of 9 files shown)


nv-yna:
## Performance validation: no measurable overhead (after the 5000ms DCGM collect-interval fix, d20d7e39)

Four same-day interleaved experiment chains on lyris GB300 (DSV4 1p3d, agentic trace replay, conc 3, 600s profiling, ~200 requests/leg; bare-`main` A legs bracketing each chain). Focus metric is **ITL p50** — within-arm spread was ≤0.4% across all chains, so ~2% effects are unambiguous.

| arm | config | ITL p50 vs same-day A | verdict |
|---|---|---|---|
| B ×2 (08-27) | tachometer 1 Hz + exporters, DCGM `--collect-interval=100` | **+1.85%** (consistent both legs) | overhead |
| C ×2 (09-06) | same but tachometer at 0.2 Hz (5s/scrape) | **+2.37%** (consistent both legs) | overhead — scrape rate irrelevant |
| D ×1 (09-06) | tachometer 1 Hz, `default_exporters: false` | **−0.76%** | parity |
| E ×1 (09-06) | tachometer 1 Hz + exporters, DCGM `--collect-interval=5000` | **+0.39%** | parity |

**Attribution:** the entire measurable cost was the DCGM exporter's 100ms (10 Hz) NVML collect interval, inherited from the power-telemetry template. Scrape frequency (B vs C), the scraper itself, and the node exporter (present in E) are all exonerated. Arm E is config-identical to this PR's amended defaults, so the shipped configuration has a directly measured leg at parity. Throughput deltas at parity arms: total tok/s −0.4% (E), tok/s/user +1.3% (D) — noise-bounded.

Other metrics (request throughput, TTFT p50/p99) showed within-arm spreads larger than between-arm deltas in every chain — including a 69.6s TTFT p99 outlier in a *bare-main* leg — i.e. workload noise, not telemetry effects.

**Not related to this PR:** conc-1200 14-node runs currently hang (`PyExecutor` HangDetector → MPI abort 137) on this cluster **on unmodified main with zero telemetry** as well — reproduced 3× across both arms and disjoint node sets after the Aug 24–25 gb300 fleet reboot. High-concurrency validation is blocked by that environment regression, not by this PR.

Full run ledger and forensics: `tmp_context/srt-slurm/20260826-obs-consolidation/progress.md` (workstation). Jobs: B=2818032/2818285, C=2933974/2934612, D=2936791, E=2937126, A legs 2817737/2818479/2933442/2934889/2936386/2937363.

PR #361, opened 2026-08-28
fix(telemetry): publish SA-Bench measurement window

anish-shanbhag:
## Problem this solves

SA-Bench could collect power telemetry without publishing the formal measurement-window markers expected by downstream processing, leaving otherwise successful runs with incomplete power results. Failures during the measured request phase also lost the window boundary needed to interpret partial runs. This change aligns SA-Bench's measured request phase, result timestamps, and telemetry window lifecycle.

## Summary
- publish the formal SA-Bench measurement window around the measured request gather
- record failed windows when the request phase raises
- expose matching benchmark wall-clock boundaries in result JSON
- raise the native power endpoint timeout default from 2s to 5s

## Validation
- focused telemetry/SA-Bench suite: 115 passed
- full suite: 1432 passed, 2 skipped, 6 deselected; 5 existing macOS/path-assumption failures
- real 8x B200 InferenceX run reproduced the missing-window failure before this fix

### DIFF
--- a/docs/config-reference.md
+++ b/docs/config-reference.md
@@ -1124,7 +1124,7 @@ telemetry:
 | `storage_subdir` | string | `power` | Output directory below the run log directory |
 | `required` | bool | `false` | Fail the benchmark when publishable power artifacts cannot be produced |
 | `startup_timeout_seconds` | float | `30.0` | Exporter readiness timeout |
-| `request_timeout_seconds` | float | `2.0` | Per-request exporter timeout |
+| `request_timeout_seconds` | float | `5.0` | Per-request exporter timeout |
 | `collector_join_timeout_seconds` | float/null | `null` | Shutdown join timeout; defaults from `request_timeout_seconds` |
 
 ---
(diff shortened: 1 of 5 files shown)


anish-shanbhag:
AIHub end-to-end validation for this change succeeded through ARIA and the
consumer `inferencex_srt_slurm` workload.

- Cluster/job: `nsc-svg-slurm-1`, Slurm `1800030`, one exclusive 8xB200 node
- Workload: Qwen3.5-397B-A17B-NVFP4, TensorRT-LLM, TP8/EP8, aggregate,
  ISL/OSL 8192/1024, concurrency 4
- Result: Slurm `COMPLETED` / `0:0`; ARIA typed result `passed=true`; 40/40
  requests and zero failures
- Throughput: 4,378.96 total tok/s; 133.20 tok/s/user; 547.37 tok/s/GPU
- Native telemetry: `captured` from `srt-slurm.dcgm-power` for all eight GPUs,
  with 78 one-second in-window samples per device
- Window: publication valid on every expected GPU, and the completed native
  window exactly matches the SA-Bench result start, end, and 75.5556-second
  duration
- Artifacts: manifest, samples, completed window, result, exporter log, request
  telemetry, and Slurm accounting were all preserved; the portable 54-file set
  excludes the task-local controller environment and lock
- Provenance: the typed result reports this PR's producer commit
  `d58c3a5cd85266e77ecc1d80784a17e2502c441b`

This also exercised the important basename collision: the formal power-window
contract and SA-Bench result are both named
`results_concurrency_4_gpus_8.json`, while the consumer correctly selected the
real result under `sa-bench_*` and the window under `power/windows`.

PR #365, opened 2026-08-28
feat(benchmark): run MLPerf via the inference-endpoint client (config passthrough)

tzulingk:
#### Overview:

Runs MLPerf on srt-slurm by driving the MLPerf team's **`inference-endpoint`** client as a `custom` benchmark.

**No MLPerf-specific schema is added to srt-slurm** — `schema.py` and `postprocess_stage.py` are byte-identical to `main`. The change is three checked-in scripts, docs, and tests.

Both phases are verified end to end on a real deployment: a real MLPerf smoke-test config produces a valid performance measurement (job 518790 — 64/64 samples, exit 0, 560 tok/s, TTFT 326 ms, TPOT 12.8 ms), and accuracy runs end to end through the scorer with results parsed into the rollup (job 522387 — 4388/4388 issued and scored). See [Status](#status).

Tracked in DIS-2798.

#### Details:

**Why `benchmark.type: custom` rather than a first-class `mlperf` type.**
Earlier revisions of this branch did make it a first-class type. That required `schema.py`
changes plus a `benchmark.rollup` field and a `postprocess_stage.py` change, it drew review
pushback, and the type kept having to grow to cover parts of the client config it did not
model. Those changes are reverted, which is why this branch has zero core changes. The
trade-off is that validation happens in-container at benchmark time rather than at
`srtctl apply`; see the discussion thread.

**Why the client config is passed through rather than re-modelled.**
It carries ~60 nested settings — model params, datasets with accuracy scoring, load pattern, transport, drain/warmup/early-stopping — and its shape moves with the client version. Re-expressing it as srt-slurm settings would be a losing race *and* lossy: anything not modelled becomes unsettable.

So the script rewrites exactly two values, the only two that cannot be known before Slurm assigns nodes:

| Rewritten | Why |
|---|---|
| `endpoint_config.endpoints` | frontend addresses are assigned at run time |
| `report_dir` | so results land with the job's logs and get collected |

Everything else passes through untouched, including unresolved `${VAR}` placeholders that the client expands itself. This mirrors the MLPerf team's own launcher (`endpoints-launch`, `NVIDIA/src/sflow/tools/generate_endpoint_yaml.py`), which rewrites one key for the same reason.

**The rollup is unified, not reinvented.**
Review asked why an earlier revision added a rollup concept. It no longer does — the `benchmark.rollup` schema field and the `postprocess_stage.py` change are both reverted. The script writes `benchmark-rollup.json` directly, the file `postprocess` already reads, using **the same flat run-record keys sa-bench emits** (`throughput_toks`, `ttft_mean_ms`, `tpot_p99_ms`, …).

That matters concretely: `cli/monitor.py::_format_run` reads those keys off the top level of each run, so a rollup that only nested its numbers would be collected and then rendered as a blank line. MLPerf-specific values with no counterpart — accuracy, the qps window, the completion verdict — live in a nested `metrics` block alongside.

**`rollup.py` from the previous revision is deleted.** It parsed LoadGen's summary, and this client does not use LoadGen; it has its own report format, now handled by `parse_report.py`.

**Two earlier revisions targeted different harnesses.** First the mlcommons reference implementations, then NVIDIA's `nv_mlpinf`. The MLPerf team has since confirmed **`inference-endpoint` is the client they will use, and `nv_mlpinf` is deprecated — v6.1 was the last round.**

**Usage:**

```yaml
benchmark:
  type: custom
  command: bash /srtctl-benchmarks/mlperf/bench.sh
  container_image: ".../endpoint_client_*.sqsh"   # client ships pre-installed here
  env:
    MLPERF_CLIENT_CONFIG: /mlperf-configs/client.yaml
    MLPERF_MODE: both        # performance and accuracy in one pass

extra_mount:
  - "/path/to/ie-configs:/mlperf-configs"    # NOT /configs - srt-slurm uses that
```

Start from `src/inference_endpoint/config/templates/submission_template.yaml` in the client repo, or one of the ~45 point configs in `endpoints-launch`.

#### Where should the reviewer start?

- `scripts/mlperf/resolve_config.py` — the two rewrites, and what is deliberately *not* touched.
- `scripts/mlperf/parse_report.py` — `canonical_run_metrics`, where this rollup unifies with sa-bench's.
- `tests/fixtures/mlperf_report/README.md` — what the fixtures are and which runs they came from.
- `tests/test_mlperf.py::TestBenchTargetsTheInjectedFrontend` — the cross-node contract, executed rather than asserted by substring.

##### Verified on the cluster

| Verified | Evidence |
|---|---|
| **A real MLPerf config runs end to end and yields a valid measurement** | Job 518790 — see below |
| **Accuracy mode runs through the scorer and lands in the rollup** | Job 522387 — 4388/4388 issued and scored, both phases parsed |
| Full orchestration → `custom` step → rollup collected | Job 514983 |
| `${VAR}` placeholders survive the rewrite and are expanded by the client | 518790 — `tokenizer_name: ${MODEL_DIR}` passed through, resolved from the recipe env |
| `MLPERF_MODE` is validated up front | 518783 spent a full bringup to learn `performance` is not a mode name; the client's are `perf`/`acc`/`both` |
| `resolve_config.py` runs and fills in the assigned frontend | 518559 — `client_config_resolved.yaml`, rest of the config preserved |
| The client accepts the resolved config | 518559 — it ran from it |
| `model_params.name` matches what the frontend serves | 518559 — requests routed and generation started; a mismatch returns 404 |
| `parse_report.py` parses a report *this PR generated* | 518559 — `wrote /logs/benchmark-rollup.json from result_summary.json` |
| Report layout search is load-bearing | 518559 wrote `result_summary.json` at the **top level**, not `performance/`. A parser fixed to the newer layout would have yielded zero metrics |
| Rollup still written when the client exits non-zero | 518559 exited 1; rollup written with the metrics it had |
| `benchmark.container_image` switches images | 516976 resolved `/opt/venv/bin/inference-endpoint`, which exists only in the client image |
| Client binary + interpreter live in a venv off `PATH` | 516954 / 517709 |
| Client configs must not mount at `/configs` | 516845 died in setup with "NATS binary not found". Documented, plus a guard in the script |

##### Verified locally against real data (44 tests)

| Verified | Against |
|---|---|
| Passthrough rewrites only the two keys | A real DeepSeek-R1 submission config; the test strips those keys and asserts the remainder is unchanged |
| Report parsing: qps/tps, counts, duration, latency blocks, accuracy + per-subset scores | Unedited output from real runs 474440 and 412798, cross-checked against the `report.txt` the client wrote beside each |
| ns → ms conversion; token counts left unscaled | Same fixtures |
| `srtctl monitor` renders the rollup | Driven through monitor's own `_rollup_runs` + `_format_run` |
| Degradation: missing report, truncated JSON, older layout, partial summary | Constructed cases |

Two fields came from the client's spec (`docs/metrics/report_design.md`) rather than the artifacts: `input_sequence_lengths`, documented but empty in every fixture here, and `legacy_loadgen_window_duration_ns`, which records *which window* produced qps/tps. Runs measured under different windows are not comparable, so the rollup carries that discriminator rather than a bare number.

##### A valid measurement (job 518790)

Run with a real MLPerf smoke-test config from `endpoints-launch`, copied verbatim apart from `model_params.name`, which must match the recipe's `served_model_name`.

```
Total samples issued: 64      Total samples failed: 0
Duration: 168.28 seconds      QPS: 0.38      TPS: 560.13
```

| | |
|---|---|
| exit code | 0, `state: complete` |
| TTFT mean / p99 | 326 ms / 1848 ms |
| TPOT mean / p99 | 12.8 ms / 19.8 ms |
| OSL mean / max | 1473 / 2000 tokens (`max_new_tokens: 2000`) |

The resolved config confirms the passthrough contract on a config nobody here wrote:

```yaml
tokenizer_name: ${MODEL_DIR}     # preserved unexpanded; the client expands it
streaming: 'on'                  # untouched
target_concurrency: 8            # untouched
- http://10.66.27.33:8000        # rewritten
report_dir: /logs/mlperf         # rewritten
```

##### An earlier run crashed; cause not attributed

Job 518559 produced no usable measurement (`osl total = 1`, every response empty). The worker log shows the engine dying:

```
[TRT-LLM] [RANK 2] Encountered an error in sampling:
                   CUDA error: unspecified launch failure
[TRT-LLM] [RANK 2] Fatal error detected, initiating shutdown
  → PyExecutor has already been shutdown   (all later requests)
  → worker stream ends with no final marker
  → Dynamo: "Disconnected: Stream ended before generation completed"
  → HTTP 500 "Failed to fold completions stream"
```

That run differed from 518790 in three ways at once — a non-MLPerf serving image, a 200-way `max_throughput` load against `max_batch_size: 32`, and non-streaming — so **the cause is not isolated**. It is recorded here because the client only ever sees the HTTP 500, in which an engine crash and a genuine response-assembly bug are indistinguishable. Surfacing the underlying `ErrorType::Disconnected` to the client would be a worthwhile Dynamo improvement, filed separately; it is not the cause of this failure.

##### Accuracy mode (job 522387)

`MLPERF_MODE: both` runs both phases and both land in the rollup:

```
exit_code: 0
metrics_sources: ["performance/result_summary.json", "accuracy/accuracy_results.json"]
response_counts:  {issued: 4388, scored: 4388, empty: 0, missing: 0}
```

This needs a client image carrying the eval subproject. `endpoint_client_bbc86973_pr372` does **not**: its installed package has only `evaluation/{extractor, livecodebench, scoring}`, and the DeepSeek runner exists solely in an uninstalled source copy, which is the `FileNotFoundError: deepseek_eval_runner.py not found` seen on an earlier run. `endpoint_client_2da91f59_vbench` installs the client editable from `/mnt/inference-endpoint/src`, where `evaluation/legacy_mlperf_deepseek_r1/` ships with its `.venv` already synced.

**Image and config cannot be chosen independently:** `eval_method` must name a scorer the image actually has. Some point configs ask for `deepseek_r1` while the image used here provides `legacy_mlperf_deepseek_r1`; they have to be picked together.

**The accuracy scores from that run are not submission-comparable** (52.29 vs 81.43 on the reference run), for two reasons, both properties of the smoke config rather than of this integration:

- `max_new_tokens` was capped at 2000 against the reference's 20000. DeepSeek-R1 is a reasoning model; `osl` max sits exactly on the cap, so most responses were truncated mid-reasoning. AIME falls furthest (6.65 vs 84.12).
- `livecodebench` is absent — 4 subsets rather than 5 — because that subset needs a separate `lcb-service` container this recipe does not deploy. That is what sets `complete: false` in the report, and the rollup preserves that flag.

##### Scope

TRT-LLM. vLLM and SGLang are out.

##### Known gap needing its own change

`MLPERF_ENDPOINTS` accepts a list because the client load-balances across frontends itself — that is how MLPerf gets past the ~28k-connection ceiling of a single `ip:port`, its own configs asking for 84,000. srt-slurm exposes one frontend, and `enable_multiple_frontends` puts nginx in front, which is again one address. That override is currently the only route at submission scale; closing it properly is srt-slurm work for a separate PR.

Rebased onto `main` now that #372 has merged. #372 is no longer a dependency — it was needed because `nv_mlpinf` hardcoded the model name, while this client takes it in its config.


richardhuo-nv:
why we are choosing the "custom" type for the benchmark? I feel this is probably too "wide-open", so that the user could very easy to forget attaching some important params or env vars.

Probably use a specific type so that we can help user to setup

```
benchmark:
  type: mlperf  # automatically invoke bash /srtctl-benchmarks/mlperf/bench.sh
  client_config_path: /path/to/ie-configs/client.yaml  # automatically mount to /mlperf-config/client.yaml or just copy to the /logs dir
  env: 
        MLPERF_MODE: both
  container_image: endpoint_client_*.sqsh  # fail fast at the srtctl apply stage if not present
```

peilii:
How is client<->frontend connectivity handled when they land on different Slurm nodes? `bench.sh` builds `ENDPOINTS` from `SRT_FRONTEND_HOST`/`SRT_FRONTEND_PORT`, which is the same injection `agentperf/bench.sh` and `benchmark_stage.py::_get_benchmark_env` use for exactly that case (client on `benchmark.client_dedicated_node` / `client_placement` != the frontend's node) - but I don't see it covered by a test here, only the same-node/local defaults.

Could we add a test (or two) around this, covering both the `perf` and `acc` modes, since `MLPERF_MODE` changes which report files get parsed and rolled up? e.g. something like:

```python
class TestBenchTargetsTheInjectedFrontend:
    """Client and frontend commonly land on different Slurm nodes
    (client_dedicated_node / client_placement != orchestrator_placement).
    SRT_FRONTEND_HOST/PORT is the real, routable frontend IP injected for
    that case - assert bench.sh's ENDPOINTS default resolves to it, not
    localhost, for both perf and acc modes."""

    def _endpoints_for(self, **env_overrides):
        env = {k: v for k, v in os.environ.items() if not k.startswith(("SRT_FRONTEND", "MLPERF_"))}
        env.update(env_overrides)
        line = next(l.strip() for l in BENCH.read_text().splitlines() if l.strip().startswith("ENDPOINTS="))
        result = subprocess.run(
            ["bash", "-c", f'{line}; printf "%s" "$ENDPOINTS"'],
            env=env, capture_output=True, text=True, check=True,
        )
        return result.stdout

    @pytest.mark.parametrize("mode", ["perf", "acc"])
    def test_targets_the_injected_frontend_node(self, mode):
        endpoints = self._endpoints_for(
            SRT_FRONTEND_HOST="10.65.4.12", SRT_FRONTEND_PORT="9000", MLPERF_MODE=mode,
        )
        assert endpoints == "http://10.65.4.12:9000"
```

That pins the cross-node contract down at the unit-test level (client reaches the frontend's real address, for either mode) rather than relying only on the hardware runs in the PR description to catch a regression here.

tests/fixtures/mlperf_client_submission.yaml
```diff
@@ -0,0 +1,78 @@
...
+    path: /work/dataset/ds-r1/deepseek_r1_eval.parquet
+    accuracy_config:
+      eval_method: legacy_mlperf_deepseek_r1
+      ground_truth: ground_truth
+      num_repeats: 1
+      extras:
+        subset_column: dataset
+        question_column: question
+        # CHANGED 2026-08-04: was the trtllm-serve path
+        #   /lustre/fsw/coreai_mlperf_inference/bofengl/models/DeepSeek-R1-NVFP4-v2-mlpinf
```
peilii: We may not want internal cluster path here (`/lustre/fsw/coreai_mlperf_inference/bofengl/...`) - it's only kept for historical context in the comment, but it also carries an employee username. Could this be genericized, e.g. `/lustre/fsw/<internal-path>/models/...` or dropped in favor of just describing what went wrong (missing local mount -> HF fallback -> HFValidationError) without the literal path?

tests/fixtures/mlperf_report/README.md
```diff
@@ -0,0 +1,24 @@
+# inference-endpoint report fixtures
+
+Unedited output from real MLPerf runs, copied from `endpoints-launch/sflow_output`
+on the hecate cluster. Kept verbatim so the parser is tested against what the
```
peilii: Same concern here - "on the hecate cluster" names an internal cluster by hostname. Suggest genericizing to something like "on an internal cluster" unless that name is already fine to have public.

peilii (commented):
Two spots carry real internal-cluster identifiers into the repo history/fixtures, worth scrubbing before this lands on a public repo:

tzulingk:
Fair question, and the answer is that we tried it your way first.

Earlier revisions of this branch did make `mlperf` a first-class benchmark type. That meant changes to `schema.py` plus a `benchmark.rollup` field and a `postprocess_stage.py` change to go with it, and it didn't work out well — reviewers pushed back on the rollup additions specifically, and the type had to keep growing to cover the parts of the client config it didn't model. Those changes are reverted here, which is why the branch is back to zero core changes: `schema.py` and `postprocess_stage.py` are byte-identical to `main`.

The underlying problem is that the client config isn't ours. It has ~60 nested settings and its shape moves with the client version, so anything srt-slurm models is a value it has to keep chasing, and anything it *doesn't* model becomes unsettable by users. `custom` sidesteps that: srt-slurm hands the client an address and a config path, and stays out of the rest.

**Your fail-fast concern is legitimate though, and I don't want to wave it away.** The script does validate, and each check exists because a real run died on it:

- `MLPERF_CLIENT_CONFIG` unset or not present in the container → hard fail
- `MLPERF_MODE` not one of `perf`/`acc`/`both` → hard fail listing the valid values (a bad mode otherwise surfaces as a bare `Required: --mode` from the client's CLI parser)
- client binary not found → hard fail naming the paths searched and the image it expects
- `SRT_FRONTEND_HOST` unset → hard fail, no localhost default, because silently benchmarking nothing is worse
- config mounted under `/configs` → warning, since that shadows srt-slurm's own `nats-server`/`etcd` and the job dies in head-node setup

But you're right that all of that happens **in-container, at benchmark time** — after allocation, model load, and the health gate. On a large deployment that's a lot of GPU-minutes to spend before learning you forgot an env var. A first-class type could reject the same mistakes at `srtctl apply`, which is strictly better for that class of error.

So the trade-off as I see it: `custom` avoids modelling a config we don't own, at the cost of late validation. If we want the early validation without the modelling, a middle option would be a thin `mlperf` type that validates *only* the things srt-slurm genuinely owns — client config path exists, `container_image` set, mode valid — and still passes the client config through untouched. That would get your fail-fast at `apply` time without srt-slurm growing opinions about the client's settings.

Happy to do that if you think it's worth it — I'd want to keep the passthrough either way, since that's the part that stops this from rotting when the client version moves.

tzulingk:
Both addressed in `5c9d8c4b`, thanks — the internal-identifier one especially, that shouldn't have been in there.

**Internal identifiers.** Scrubbed the two you flagged and swept the rest of the diff for the same class of thing:

- `mlperf_client_submission.yaml` — dropped the cluster path and username; the comment now describes the failure (unmounted path → HF fallback → `HFValidationError`) without the literal path. Also removed the hardware/topology identifiers from the header and the config name, and genericized `tokenizer_path` to `/models/deepseek-r1`. It is otherwise still the real config, and the header now says explicitly that paths and identifiers were genericized so nobody mistakes it for byte-verbatim.
- `mlperf_report/README.md` — "on an internal cluster", and the run rows now carry just the run numbers.
- Also caught in the sweep: hardware identifiers in `test_mlperf.py`'s fixture comment and in `parse_report.py`'s docstring, which named specific runs and generations.

I left the `endpoints-launch` repo references, since they point at where a user finds their client config rather than at a machine or a person — say the word if you'd rather those went too.

**Cross-node test.** Added, close to your sketch. One deliberate difference: it runs the `ENDPOINTS` line under `bash` with the environment varied, rather than asserting on the script text, so it fails if the *behaviour* changes and not merely the wording:

```python
@pytest.mark.parametrize("mode", ["perf", "acc", "both"])
def test_targets_the_injected_frontend_node(self, mode):
    assert self._endpoints_for(
        SRT_FRONTEND_HOST="10.65.4.12", SRT_FRONTEND_PORT="9000", MLPERF_MODE=mode,
    ) == "http://10.65.4.12:9000"
```

Plus two neighbours the same seam needs: the port defaults to 8000 while the host deliberately has no default (a localhost fallback would benchmark nothing and still exit 0), and `MLPERF_ENDPOINTS` overrides the injected value — that's the multi-frontend escape hatch, and it has to win.

I mutated `ENDPOINTS` to `http://localhost:...` to check the tests actually bite: 4 of them fail. Without that check I'd only have known they pass.

Added `both` alongside `perf`/`acc` since that's the default mode and the one most runs will use. 44 tests total.

tzulingk:
Ran this on real nodes as well, since the unit tests pin the contract but had never put the client on a different machine from the frontend.

Same recipe as the verified run, with only the placement changed:

```yaml
benchmark:
  type: custom
  client_dedicated_node: true
  client_placement: head
```

**First attempt was inconclusive, which is worth mentioning.** The job completed and the client used the frontend's routable IP — but nothing in the run artifacts recorded the client's *own* node, and `SRT_FRONTEND_HOST` resolves to the frontend whether or not the client is elsewhere. So a successful cross-node run and a successful same-node run were indistinguishable after the fact. I could have reported it as proof; it wasn't.

That gap belonged to the script rather than the run, so `bench.sh` now logs the client's node next to the address it is targeting:

```bash
echo "[mlperf] client node: $(hostname) -> targeting $ENDPOINTS"
```

Re-running gives the evidence directly:

```
[mlperf] client node: hecate0123 -> targeting http://10.66.27.97:8000

hecate0121 -> 10.66.27.97     # frontend + worker
hecate0123 -> 10.66.27.101    # client
```

Client on one node, reaching the frontend on another across the network; the client's own address appears nowhere in the endpoint. `exit 0`, 64/64 samples, 556.3 tok/s — in line with the same-node run's 560.1.

Every run now carries this line, so placement is checkable from the logs rather than inferred from the recipe.

**Not covered:** `client_placement: last_decode` on a disaggregated deployment. That is a different code path from `client_dedicated_node` and this run says nothing about it.

nv-yna (approved):

nv-yna:
would be great to add a mlperf GB300 recipe for reference and example

peilii (approved):

tzulingk:
@nv-yna added in `238f1201` — `recipes/gb300-fp4/mlperf-dsr1-agg.yaml`.

It is the aggregated DeepSeek-R1 NVFP4 configuration that was validated on hardware, with the paths reduced to placeholders. The comments carry the four things that are easy to get wrong and each cost a real run to find:

- `served_model_name` has to equal the client config's `model_params.name`, or every request 404s
- client configs must **not** mount at `/configs` — srt-slurm keeps its own `nats-server` and `etcd` binaries there, and shadowing them kills the job during head-node setup
- `MLPERF_MODE` takes the client's own names (`perf`/`acc`/`both`), not the `performance`/`accuracy` spellings the config uses for dataset *types*
- the client image and the client config have to be chosen together: accuracy needs an image with the eval subproject installed, and `eval_method` must name a scorer that image actually ships

Validates under `validate-recipes` along with the other 555.

Also resolved the two threads from @peilii — the scrub landed in `5c9d8c4b`, I had just left them open in the UI.

PR #370, opened 2026-09-01
feat(infra): make NATS/etcd startup timeout configurable

bigsiu:
## Problem

`start_head_infrastructure` waits a fixed 300s for the NATS port (and etcd), but the budget starts before the workload container exists on the infra node. On a cold enroot cache, the first Pyxis import of a multi-GB registry image counts against that budget and can exceed it on its own. When the timeout fires, `infra.out` contains only `pyxis: importing docker image: ...` and the sweep dies with `RuntimeError: NATS failed to start`, although NATS itself is healthy.

Measured across repeated runs on GB200/GB300 NVL72 Slurm clusters: imports take ~143-161s on warm-cache nodes and exceed 300s on cold nodes (one observed import took over an hour), so the same recipe passes or fails purely by per-node cache state. A controlled A/B with the repo's own smoke test (`recipes/mocker/agg.yaml`) reproduces it deterministically: green on a warm node, dead at exactly the 300s mark on a cold one. Recipes using pre-imported local `.sqsh` files never see this, which is likely why it went unreported.

## Change

1. New `infra.startup_timeout` config field (default 300, stock behavior unchanged), documented as including container import time for registry-URI containers.
2. The NATS and etcd port waits in `do_sweep.py` use it instead of the hardcoded constant.

Minimal and backward compatible. Verified end to end: with the field set, previously flaky recipes complete green on cold nodes across both cluster generations.

## Follow-up (happy to build if wanted)

The deeper fix is to start the clock only once the container entrypoint is actually running (a sentinel written by `setup_head.py`), so import time never counts against service startup and no fixed number needs sizing. Our data shows any fixed timeout eventually loses to a slow-enough import; the sentinel design is the durable variant. I kept this PR minimal per the discussion, and can follow up with the sentinel as a separate PR.

### DIFF
--- a/src/srtctl/cli/do_sweep.py
+++ b/src/srtctl/cli/do_sweep.py
@@ -202,14 +202,19 @@ class SweepOrchestrator(
             critical=True,
         )
 
-        # 300s timeout to handle slow container imports on first run
+        # The wait budget must absorb the container import on the infra node:
+        # the srun above triggers the Pyxis/enroot import before setup_head.py
+        # runs, so on a cold enroot cache the first import of a multi-GB image
+        # counts against this timeout and can exceed the 300s default on its
+        # own. Configurable via infra.startup_timeout.
+        startup_timeout = self.config.infra.startup_timeout
         logger.info("Waiting for NATS (port %d) on %s...", NATS_PORT, infra_node)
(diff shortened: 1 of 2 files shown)


bigsiu:
Superseded by #371 (same change from my NVIDIA-affiliated account PiotrBigajNV).

PR #371, opened 2026-09-01
feat(infra): make NATS/etcd startup timeout configurable

PiotrBigajNV:
## Problem

`start_head_infrastructure` waits a fixed 300s for the NATS (and then etcd) port, but the budget starts before the workload container exists on the infra node. On a cold enroot cache, the first Pyxis import of a multi-GB registry image counts against the wait and can exceed 300s on its own, so the same recipe passes or fails purely by per-node cache state. When it fires, the job dies with `RuntimeError: NATS failed to start` while NATS itself is healthy; `infra.out` shows only `pyxis: importing docker image`.

Measured on GB200/GB300 NVL72 Slurm clusters: 143-161s import on warm-cache nodes, over 300s on cold nodes. A controlled A/B with the repo's own `recipes/mocker/agg.yaml` reproduces it: green on a warm node, dead at exactly the 300s mark on a cold one. Recipes using pre-imported local `.sqsh` files never see this.

## Change

Per maintainer feedback, the knob lives in top-level `srtslurm.yaml` (an environment characteristic, not a workload one), with a per-submission escape hatch:

1. `startup_timeout` in `srtslurm.yaml` — declared as a validated `ClusterConfig` field (must be >= 1s), used by both the NATS and etcd port waits. Stock 300s default preserved when unset. Documented in `srtslurm.yaml.example`.
2. `srtctl apply --startup-timeout N` — overrides the configured value for that submission only (carried to the sweep via `SRTCTL_STARTUP_TIMEOUT` in the generated job script), for one-off submissions whose image differs materially from the cluster norm.

Resolution order: CLI flag > `srtslurm.yaml` > stock 300s default. Recipes are unchanged.

## Tests

`tests/test_configs.py`: schema round-trip incl. rejection of 0, CLI-override-beats-yaml, yaml-beats-default. Full config suite green (206 passed).

### DIFF
--- a/src/srtctl/cli/do_sweep.py
+++ b/src/srtctl/cli/do_sweep.py
@@ -202,14 +202,27 @@ class SweepOrchestrator(
             critical=True,
         )
 
-        # 300s timeout to handle slow container imports on first run
+        # The wait budget must absorb the container import on the infra node:
+        # the srun above triggers the Pyxis/enroot import before setup_head.py
+        # runs, so on a cold enroot cache the first import of a multi-GB image
+        # counts against this timeout and can exceed the 300s default on its
+        # own. Resolution: per-submission CLI override (srtctl apply
+        # --startup-timeout, carried as SRTCTL_STARTUP_TIMEOUT by the sbatch
+        # script) > srtslurm.yaml startup_timeout > 300.
+        from srtctl.core.config import get_srtslurm_setting
+
+        env_override = os.environ.get("SRTCTL_STARTUP_TIMEOUT")
+        if env_override is not None:
+            startup_timeout = int(env_override)
+        else:
+            startup_timeout = int(get_srtslurm_setting("startup_timeout", 300))
         logger.info("Waiting for NATS (port %d) on %s...", NATS_PORT, infra_node)
-        if not wait_for_port(infra_node, NATS_PORT, timeout=300):
+        if not wait_for_port(infra_node, NATS_PORT, timeout=startup_timeout):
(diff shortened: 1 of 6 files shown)


PiotrBigajNV:
Reworked per the Slack discussion: the primary knob is now an environment-level `startup_timeout` in `srtslurm.yaml` (per-cluster import speed is an environment characteristic, not a workload one), default 300 preserved when unset. The recipe-level `infra.startup_timeout` remains as an optional per-run override for workloads whose image differs materially from the cluster norm. Tests cover the resolution order (recipe > srtslurm.yaml > 300).

PiotrBigajNV:
Reworked to srtslurm.yaml-only as suggested: `startup_timeout` is now purely an environment-level setting (default 300 when unset), recipes are untouched. PR description updated; head is `01cd693`.

ishandhanani:
Closing with the 2.0 merge (#407): etcd and NATS are now services with a 300 s readiness default, and a recipe that needs more declares them and sets `readiness.timeout_seconds`:

ishandhanani:
The example that was cut off above:

```yaml
services:
  - name: etcd
    type: etcd
    readiness:
      port: 2379
      timeout_seconds: 900
  - name: nats
    type: nats
    readiness:
      port: 4222
      timeout_seconds: 900
```

If a cluster-level knob in `srtslurm.yaml` is still wanted on top of that (the cold-cache import time is an environment property, as you noted), please reopen against `main`; the place for it now is a default on the etcd and nats service kinds. See docs/services.md.

PR #373, opened 2026-09-01
fix(profiling): preserve workers during timed nsys capture

nbommi19:
## Summary

- replace `nsys profile --delay/--duration` worker ownership with named `nsys launch/start/stop` sessions
- keep the serving worker alive after a bounded collection window so SRT-Slurm remains lifecycle owner
- use the same time-capture wrapper for vLLM, SGLang, and TensorRT-LLM
- use `cuda-sw` for delayed CUDA collection on Blackwell and export the requested report after `nsys stop`
- validate the capture window and cover session registration, shutdown, and application lifetime in tests

## Why

The previous duration-based wrapper either terminated the serving process when collection ended or allowed the surrounding Slurm/Pyxis step to tear it down when `nsys` exited. Framework HTTP profiling endpoints also differ or are unavailable. A named Nsight session makes collection independent of framework APIs and separates profiler lifetime from worker lifetime.

## Validation

- focused SRT-Slurm profiling tests: 29 passed
- `ruff` and shell syntax checks: passed
- full `make check`: 1,433 passed, 2 skipped; five existing macOS/worktree-specific failures remained (mock Slurm process behavior, sa-bench's adjacent `lib/profiling.sh` fixture, and Linux-only CPU-affinity coverage)
- integrated B200 cluster runs on `nsc-svg-slurm-1`: vLLM, SGLang, and TensorRT-LLM each completed AIPerf after the Nsight report finalized and produced a non-empty report/SQLite export with CUDA kernels

## Dependency

This commit is consumed by lights-out-inference MR !855, which adds the deterministic AIPerf profile-capture task and artifact validation.

### DIFF
--- a/src/srtctl/core/schema.py
+++ b/src/srtctl/core/schema.py
@@ -800,7 +800,7 @@ class ProfilingConfig:
 
     type: str = "none"  # "none", "nsys", "nsys-time", or "torch"
 
-    # Extra arguments passed to nsys profile (appended before `-o`; see get_nsys_prefix)
+    # Extra arguments passed to nsys profile, or to nsys start for nsys-time.
     extra_nsys_args: list[str] | None = None
 
     # Phase-specific profiling step configs (not used for nsys-time)
@@ -808,9 +808,9 @@ class ProfilingConfig:
     decode: ProfilingPhaseConfig | None = None
     aggregated: ProfilingPhaseConfig | None = None
 
-    # nsys-time fields: time-based capture window, same on all workers
-    delay_secs: int | None = None  # nsys --delay: seconds from worker launch before capture starts
-    duration_secs: int | None = None  # nsys --duration: seconds to capture after delay
+    # nsys-time fields: named-session capture window, same on all workers
+    delay_secs: int | None = None  # Seconds from worker launch before collection starts
+    duration_secs: int | None = None  # Seconds to collect before finalizing the report
     benchmark_duration_secs: int = 300  # total traffic generation duration (must cover delay + duration)
 
     @property
@@ -825,7 +825,7 @@ class ProfilingConfig:
 
     @property
     def is_nsys_time(self) -> bool:
-        """Check if using time-based nsys capture (--delay/--duration instead of cudaProfilerApi)."""
+        """Check if using a wall-clock nsys capture instead of cudaProfilerApi."""
         return self.type == "nsys-time"
 
     @property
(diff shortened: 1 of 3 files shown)


nbommi19:
Closing this because ARIA no longer requires an SRT-Slurm change.

Why I opened it: the original ARIA integration used `nsys profile --delay/--duration` around the worker. When that bounded collection returned, the worker/Slurm step could be torn down before the benchmark completed. A named `nsys launch/start/stop` lifecycle fixed that behavior consistently for vLLM, SGLang, and TensorRT-LLM, including delayed `cuda-sw` capture on B200.

Why it is no longer required: after re-auditing ARIA's exact pinned SRT-Slurm commit (`13ff0f8d875490891fcf20fcfd019afbc0e8a354`), its existing public hooks are sufficient. It already has `profiling.type: nsys-time`, applies `SRTCTL_NSYS_BIN` to the actual worker command for all three frameworks, and accepts the required worker mounts/environment. ARIA can therefore stage a task-local compatibility adapter that translates the existing command to the named-session lifecycle, while keeping this repository and ARIA's submodule pin unchanged.

The final ARIA-only implementation was validated on `nsc-svg-slurm-1` with the unchanged pin and a full 180-second AIPerf run:

- vLLM, Slurm `1842368`: passed, 1,759,568 CUDA kernel rows
- SGLang, Slurm `1842401`: passed, 2,179,759 CUDA kernel rows
- TensorRT-LLM, Slurm `1842442`: passed, 1,162,930 CUDA kernel rows

Each run produced a non-empty `.nsys-rep` and readable SQLite export, and each server continued serving after the 10-second capture until AIPerf completed. The underlying change here may still be useful as a general SRT-Slurm enhancement, but ARIA no longer depends on it, so I am closing this PR to avoid an unnecessary cross-repository dependency.

nbommi19:
Follow-up on why this was opened and why it is no longer required:

The first integration used a duration-bounded `nsys profile` process. When the
profiler returned, its worker step could end before the benchmark completed, so
the missing lifecycle initially looked like it belonged in the launcher.

After changing the requirement to exact engine iterations through
`cudaProfilerApi`, I re-audited the SRT integration surface. The existing
`profiling.type: nsys` path already exposes the framework iteration controls and
the worker-side nsys command override needed by the integration. The remaining
trigger adaptation and artifact validation can stay on the integration side.

The final integration therefore keeps its existing SRT revision and does not
consume this PR. That is why this PR is closed rather than being added as a new
dependency. Thanks for the review context while that ownership boundary was
being worked out.

PR #374, opened 2026-09-02
feat(auxiliary-services): generic sidecar config for both execution paths

ishandhanani:
## Summary

Adds a generic top-level `auxiliary_services` recipe block for declaring bolt-on sidecar processes -- persistent background services launched alongside the main inference workers -- without needing a bespoke srtctl stage every time. Same role Kubernetes' `podTemplate.spec.containers` plays: an arbitrary component wired in from the recipe (name, command, container image, env, and an optional git source to build first).

## What's new

- `AuxiliaryServiceConfig` / `AuxiliaryServiceSourceConfig` schema (`src/srtctl/core/schema.py`), with validation for uniqueness, non-empty command/name, immutable source revs, and a warning when `source` is set without `build_command`.
- `dry-run` rendering of every configured service's command/env/source (`src/srtctl/cli/submit.py`), and serialization into the direct-plan JSON (`src/srtctl/render/direct_plan.py`).
- `docs/auxiliary-services.md`, with a quick start, full config reference, and the motivating `dynamo.thunderagent_router` example.

## Two-path design

Implemented once per execution path, behind the same config:

- **Real multi-node `sbatch`/SLURM path**: `AuxiliaryServiceStageMixin` in `src/srtctl/cli/mixins/auxiliary_stage.py`, used by `SweepOrchestrator`. Every service launches via `srun` on the head node only, non-critical (a dead sidecar never tears down the benchmark) -- same placement and criticality contract as Tachometer. Crash detection and cleanup ride the shared `ProcessRegistry`.
- **`--bash` dev-mode path**: `AuxiliaryServiceStageMixin` in `src/srtctl/render/direct_stages/auxiliary_stage.py`, used by `DirectRunner`. Every service runs as a plain local subprocess; a service that exits within ~2 seconds of launch fails the whole run.

Both paths start auxiliary services once workers and the frontend are confirmed healthy, and inject `ETCD_ENDPOINTS`/`NATS_SERVER` discovery env by default (`inherit_discovery_env: true`).

## Declared-order simplification

Services launch in the order they're declared in the `auxiliary_services` YAML list -- no dependency graph. Every real use case just needs a service to launch after workers/frontend are healthy, which is already true by construction on both paths (the start method is only called once workers/frontend are confirmed ready). There's no real use case for one auxiliary service depending on *another* auxiliary service, so an earlier `depends_on` + topological-sort-with-cycle-fallback design was cut in favor of this simpler contract.

## Validation

- `make check` is clean (lint + full test suite), matching the pre-existing baseline on this environment (two unrelated failures on macOS -- `os.sched_getaffinity` not available, and an unrelated `cpu_allocation` assertion -- present identically on `origin/main` before this branch's changes).
- New coverage: config loading/validation, discovery-env injection, declared-order launching, and the clone-then-build-then-launch source sequencing, on both the dev-mode and sbatch paths.

### DIFF
--- a/docs/auxiliary-services.md
+++ b/docs/auxiliary-services.md
@@ -88,6 +88,7 @@ auxiliary_services:
       path: subdir                      # optional, subdirectory to build/launch from
     build_command: ["bash", "-lc", "pip install -e ."]  # required if source is set
     inherit_discovery_env: true         # optional, default: true
+    critical: false                     # optional, default: false
 ```
 
 | Field                    | Required | Default          | Notes                                                                 |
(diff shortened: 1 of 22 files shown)


ishandhanani:
Closing with the 2.0 merge (#407, part 11): the `services:` block landed there, carrying this PR's sidecar config and its kv_events fix.

PR #383, opened 2026-09-04
feat(postprocess): exclude AIPerf inputs.json from uploaded artifacts.

fsaady:
Feature Summary:

Exclude AIPerf-generated inputs.json files from saved artifacts. During postprocessing, srt-slurm temporarily moves these files outside the log directory so artifact rollup and upload ignore them, then restores them afterward. This applies to all benchmarks using `AIPerfBenchmarkRunner`.

Reasoning:

AIPerf now generates inputs.json, which can be large and is not useful as a persisted benchmark artifact. Keeping it increases archive size and upload/storage costs unnecessarily. Temporarily quarantining it avoids publishing the file while preserving the local output and leaving benchmark runners independent of artifact-handling concerns.

### DIFF
--- a/src/srtctl/cli/mixins/postprocess_stage.py
+++ b/src/srtctl/cli/mixins/postprocess_stage.py
@@ -66,7 +66,7 @@ class _QuarantinedArtifact:
     quarantined_path: Path
 
 
-_AIPERF_ARTIFACT_EXCLUSIONS = (_ArtifactExclusion(pattern="artifacts/**/inputs.json"),)
+_AIPERF_ARTIFACT_EXCLUSIONS = (_ArtifactExclusion(pattern="**/inputs.json"),)
 
 
 class PostProcessStageMixin:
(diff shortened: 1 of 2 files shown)


src/srtctl/cli/mixins/postprocess_stage.py
```diff
@@ -45,6 +50,25 @@
...
+
+@dataclass(frozen=True)
+class _QuarantinedArtifact:
+    """An artifact temporarily moved out of the upload tree."""
+
+    original_path: Path
+    quarantined_path: Path
+
+
+_AIPERF_ARTIFACT_EXCLUSIONS = (_ArtifactExclusion(pattern="artifacts/**/inputs.json"),)
```
fsaady: This only works because I was using --artifact-dir in the aiperf command. Making it more generic now

fsaady: Changed it to `**/inputs.json`.

Reviewers: what do you think?

Issue #385, opened 2026-09-05
[SRT-SLURM 2.0] Roadmap


ishandhanani:
2.0 babyyyyyyyyyyyyyyyyyyyyy - versioning literally means nothing for this project btw

<img width="823" height="824" alt="Image" src="https://github.com/user-attachments/assets/50671997-7939-4e4b-b3ae-09279ed778b7" />

1. Observability
    - All scraping goes through tachometer. We will remove all ad-hoc python based scraping tools. Any post processing of information into graphs/etc needs to be re-done and powered by the parquet that tachometer provides. Using python in the big 26 is pretty insane anyway
    - The current performance dashboard can be SIGNIFICANTLY improved and I'll be leveraging my insanely tuff UI duffy (fable 5.1) to improve it.
    - It is unclear how logging is applied to dynamo/sglang/etc services. It will all be tied into the observability revamp

2. Recipes
    - I'm deleting every single recipe in this repo. If you PR a recipe - auto ban (just kidding I don't know how to do this)

3. Code cleanups
    - All of my single node/bash work needs to be stripped out and plugin-ifed
    - Mooncake store cleanup https://github.com/NVIDIA/srt-slurm/pull/265
    - Docs need to be refreshed with updated schema. Need CI checks to make sure docs are updated
    - I get the vibe that our nsys/profiling stuff rn is a bit wack...need to investigate

4. Features
    - Auxiliary container deployments - https://github.com/NVIDIA/srt-slurm/pull/374
    - vllm router - https://github.com/NVIDIA/srt-slurm/pull/340
    - Overall improvement in how caching works for containers/wheels/rust builds. There are a few PRs up for this but it needs to be improved a good bit so its faster
    - MCP improvements

ishandhanani:
## 2.0 execution plan

This consolidates the roadmap above into tracks, after auditing the open PR board, the power telemetry series, and the current schema. Nothing here has started; this is the order of attack.

### Ground rules

- **Power telemetry is frozen.** `src/srtctl/core/power/`, the power hooks in the telemetry and benchmark stages, `measurement_window.py`, the `telemetry:` YAML block, and the in-flight CPU-power work (#381 and the branches it builds on) are not touched by any task below. That area is still in flux downstream and gets its own track later.
- **The v1 recipe loader stays on `main` indefinitely.** InferenceMAX carries several hundred srt-slurm recipes and many of its lanes clone `main` directly. Anything that breaks v1 loading breaks production the day it merges.
[...]

PR #386, opened 2026-09-06
[2.0] Remove the --bash direct-host execution path

ishandhanani:
## Summary

First PR of the 2.0 stack (plan: #385, Track 1 item 4). Removes the `--bash` direct-host lifecycle so 2.0 is the Slurm path only. Pure deletion plus the CLI flag and doc references; no behavior change for `srtctl apply`, `dry-run`, or any Slurm stage.

Deleted: `src/srtctl/render/` (direct plan, host runner, container runner, direct stage mixins), `templates/direct_container.sh.j2`, `docs/direct-host.md`, the three `test_direct_*` suites.

Edited: `cli/submit.py` (drop `--bash`, `render_bash_script`, and the bash-mode argument checks), `tests/test_submit_cli.py` (the direct-container test becomes a check that `--bash` is rejected), five docs, and three stale comments.

Untouched: everything under `core/power/`, the power hooks, `measurement_window.py`.

## Validation

- `ruff check` and `ruff format --check` clean
- `pytest tests`: 1656 passed, 2 skipped, 6 deselected
- Cluster smoke on this branch: PASSED. Job 11833 on sa-b200 (Qwen3-0.6B, 2x TP1 SGLang behind the SGLang router, sa-bench isl 128 osl 128 c4): workers healthy, benchmark ran, rollup written, clean teardown. Same recipe passed on main as job 11830 for a baseline.

## Stack

1. **this PR** remove `--bash`
2. recipes to a curated `examples/` matrix
3. schema-generated config reference with CI drift check


ishandhanani:

PR #388, opened 2026-09-06
[2.0] Generate docs/schema-reference.md from the config dataclasses

ishandhanani:
## Summary

Third PR of the 2.0 stack (plan: #385, Track 1 item 2). Stacked on #387; the diff against that branch is what to review.

`docs/config-reference.md` is hand-written prose and drifts from the schema (the power docs merged last week already reference a `telemetry.provider` field that #317 removed). This adds a generated, CI-checked field-level reference so keys, types, and defaults live in the code.

- `src/srtctl/core/schema_docs.py` walks the dataclass tree rooted at `SrtConfig`, the four backend types, and `ClusterConfig`, and renders one Markdown table per dataclass: YAML key (honoring marshmallow `data_key`, so `gpus_per_prefill` rather than `_explicit_gpus_per_prefill`), type (nested dataclasses linked, `Literal`s expanded), default, and a description taken from the class docstring `Attributes:` block or the `#` comment on the field.
- `srtctl schema-docs` writes `docs/schema-reference.md` (562 lines, 38 sections). `srtctl schema-docs --check` exits 1 when the checked-in file is stale.
- Enforcement in three places: a CI lint step, `make schema-docs-check` (part of `make check`), and `tests/test_schema_docs.py::test_checked_in_schema_reference_is_current`.
- `docs/config-reference.md` now points at the generated file as authoritative for keys, types, and defaults; `SUMMARY.md` links it; `CLAUDE.md` tells contributors to regenerate after any schema change and where to put field descriptions so they land in the table.

## Validation

- `ruff`, `ruff format --check`, and `ty` clean on the new module
- `tests/test_schema_docs.py` (9 tests: drift check, determinism, data_key handling, docstring and comment descriptions, backend and cluster sections, CLI write/check/stale)
- `srtctl schema-docs --check` passes on the committed file

## Stack

1. #386 remove `--bash`
2. #387 examples matrix
3. **this PR** schema-generated reference

PR #387, opened 2026-09-06
[2.0] Replace the recipes archive with a curated examples matrix

ishandhanani:
## Summary

Second PR of the 2.0 stack (plan: #385, Track 1 item 1). Stacked on #386; the diff against that branch is what to review.

Removes `recipes/` (563 files) and replaces it with `examples/`, one small runnable configuration per frontend and topology. Every example serves Qwen3-0.6B on one node, so the files differ only in the frontend and the prefill/decode layout and a full matrix run finishes in minutes.

```
examples/
  sglang/   dynamo-agg  dynamo-disagg  sglang-router-agg  sglang-router-disagg
  vllm/     dynamo-agg  dynamo-disagg  vllm-router-agg    vllm-router-disagg   vllm-direct-agg
  trtllm/   dynamo-agg  dynamo-disagg  trtllm-serve-agg   trtllm-serve-disagg
  mocker/   dynamo-agg
  features/ sweep  override  profiling
```

`examples/README.md` documents the matrix, the two `srtslurm.yaml` alias kinds the files rely on (`qwen3-0.6b`; `sglang`, `vllm`, `trtllm`), and how to validate.

Supersedes #338 and carries its wiring (CI validate step, Makefile target, interactive selector, CODEOWNERS, docs, tests).

## Code change

`validate_config_file` now expands top-level `sweep:` files the same way it expands `base:`/override files, so CI validates every file under `examples/`. Previously sweep files were rejected as having an unknown `sweep` field.

## Validation

- All 17 examples pass `validate_config_file` (sweep and override files expanded to their variants)
- `ruff` clean on `src/srtctl/`
- Targeted suites green: configs, mocker, integration status, e2e, submit CLI, dry-run, sweep, override
- Cluster: pending. Will run the matrix on sa-b200 (Qwen3-0.6B) and record per-example results here.

## Stack

1. #386 remove `--bash`
2. **this PR** examples matrix
3. schema-generated config reference with CI drift check

PR #389, opened 2026-09-06
[2.0] Add the recipe schema version key and srtctl migrate

ishandhanani:
## Summary

Fourth PR of the 2.0 stack (plan: #385, Track 2 step 1). Stacked on #388; the diff against that branch is what to review.

Recipes gain a top-level `schema:` key. Absent means `1` (the pre-2.0 layout); `schema: 2` is the 2.0 layout. Both versions load on `main`; unknown versions are rejected at load time. This is the dispatch point the structural 2.0 changes (`roles:`, `placement:`, `services:`) hang off, and it lets downstream recipes declare which layout they target instead of failing on unknown fields.

- `SrtConfig.schema_version` (YAML key `schema` via marshmallow `data_key`), plus `CURRENT_SCHEMA_VERSION` and `SUPPORTED_SCHEMA_VERSIONS`.
- Override files declare `schema:` beside `base:`; `generate_override_configs` and `resolve_override_yaml` carry it into every expanded variant. Sweep files carry it through expansion unchanged.
- `srtctl migrate -f recipe.yaml [--in-place | --output PATH]`: a ruamel round-trip migration that inserts or bumps `schema: 2` and preserves comments, key order, and quoting. Structural v1 to v2 rewrites plug into `_migrate_1_to_2` in later steps.
- All 17 examples were migrated with the new command (dogfooding), so `examples/` is the v2 reference corpus.
- Docs: `config-reference.md` gains a `## schema` section; `schema-reference.md` regenerated (the new field shows up automatically).

## Validation

- `ruff` clean; targeted suites 391 passed; full suite green
- `tests/test_schema_version.py` (14 tests): absent/2/unknown versions, propagation through override variants, validation of override and sweep files with the key, migration idempotence and comment preservation, override and lock files keep top-level sections, CLI stdout / in-place / output modes, every example declares the current version

## Stack

1. #386 remove `--bash`
2. #387 examples matrix
4. **this PR** schema version key and migrate

### DIFF
--- a/docs/config-reference.md
+++ b/docs/config-reference.md
@@ -148,6 +148,21 @@ This is useful for portable recipes that you want to share across clusters or ha
 
 ---
 
+## schema
+
+| Field    | Type    | Required | Description                                                                 |
+| -------- | ------- | -------- | --------------------------------------------------------------------------- |
+| `schema` | integer | No       | Recipe schema version. Absent means `1` (the pre-2.0 layout); `2` is current. |
+
+Every supported version loads. Put the key first in the file, beside `base:` in an override file. Upgrade a recipe with `srtctl migrate -f recipe.yaml --in-place`, which preserves comments and key order.
+
+```yaml
+schema: 2
+name: "deepseek-r1-benchmark"
+```
+
+---
+
 ## name
 
 | Field  | Type   | Required | Description                                        |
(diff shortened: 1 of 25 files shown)


ishandhanani:

PR #390, opened 2026-09-06
[2.0] Add --set KEY=VALUE and --unset KEY recipe overrides

ishandhanani:
## Summary

Fifth PR of the 2.0 stack (plan: #385, Track 2 step 2). Stacked on #389; the diff against that branch is what to review.

Downstream runners edit recipes with `sed` anchored on v1 indentation and field names, which breaks silently under any re-nesting. This adds a supported, path-based override so scripts can tweak a recipe without editing the YAML, before the structural 2.0 changes land.

```bash
srtctl apply -f recipe.yaml --set health_check.max_attempts=720 --unset sbatch_directives.exclude
srtctl apply -f recipe.yaml --set 'backend.sglang_config.decode.speculative-config={"method": "eagle"}'
srtctl dry-run -f recipe.yaml --set benchmark.concurrencies=[4,8]
```

- Paths are dotted, `[N]` indexes a list, quotes protect segments with dots. Values parse as YAML scalars or lists; mappings stay literal strings because engine flags take JSON.
- Applied in `materialize_config_path` to the raw document (ruamel round trip, comments kept) before cluster defaults, observability expansion, sweep expansion, and validation, so an explicit `--set` always wins and `{placeholder}` values still expand.
- On override files the value is written into `base` and every `override_*` / `zip_override_*` variant (one-element-list broadcast for zip groups), so no variant can shadow it. `--unset` removes from all of them and is a no-op on missing paths.
- The overridden document is what becomes `config.yaml` in the job directory; the source file is never modified. Every `--json` record lists `applied_overrides`.
- Available on `apply`, `dry-run`, `preflight`, and `resolve-override`.

## Validation

- `ruff` clean; full suite green
- `tests/test_overrides.py` (24 tests): path grammar and rejections, value typing incl. JSON-as-string and placeholders, nested creation, list indexes, lenient unset, override-file semantics, comment preservation, dry-run, `--set` beating `default_health_check`, `--json` record and written `config.yaml` under `--mock`, `resolve-override --stdout`
- Cluster: PASSED on sa-b200. Job 11866 ran examples/sglang/sglang-router-agg.yaml with --set name=..., --set benchmark.concurrencies=[2], --set backend.sglang_config.aggregated.max-running-requests=32, --unset slurm.time_limit. The written outputs/11866/config.yaml showed name renamed, concurrencies=[2], max-running-requests=32; the sbatch script fell back to the cluster default --time=4:00:00 after the unset; the --json record listed all four applied_overrides.

## Stack

1. #386 remove `--bash`
2. #387 examples matrix
5. **this PR** `--set` / `--unset`

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -254,13 +254,22 @@ srtctl apply -f <config.yaml> [options]
 | `--setup-script` | Custom setup script from `configs/` |
 | `--tags` | Comma-separated tags for the run |
 | `--serve-only` | Deploy the endpoint without running a benchmark; serve until cancellation |
+| `--set KEY=VALUE` | Override one recipe value by dotted path before validation (repeatable). Also on `dry-run`, `preflight`, `resolve-override` |
+| `--unset KEY` | Remove one recipe key by dotted path before validation (repeatable) |
 | `-y, --yes` | Skip confirmation prompts |
+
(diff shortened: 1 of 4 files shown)


ishandhanani:

PR #391, opened 2026-09-06
[2.0] Resolve every container alias with one recipe walker

ishandhanani:
## Summary

Sixth PR of the 2.0 stack (plan: #385, Track 2 step 3). Stacked on #390; the diff against that branch is what to review.

Container-alias resolution was seven copy-pasted blocks in `resolve_config_with_defaults`, one per image key (`model.container`, `frontend.container_image`, `frontend.nginx_container`, `benchmark.container_image`, the Tachometer and power exporter images), and `backend.mooncake_kv_store.container` had no block at all. This replaces them with one walk over the recipe.

- Any string under a `container`, `container_image`, `image`, or `nginx_container` key that names a `containers:` alias is resolved; literal paths and registry URIs pass through.
- Free-form maps (`environment`, `*_environment`, `env`, `args`, engine config blocks, `container_mounts`) and the `identity` block are skipped, so a key that happens to be named `image` in user data is never rewritten.
- A new recipe block that names an image (e.g. the future `services:` list) resolves with no resolver code. This is the seam Track 3 and step 7 need.

## Validation

- `ruff` clean; full suite: 1759 passed
- `tests/test_container_aliases.py`: every image key in one pass (incl. the newly-covered Mooncake container), the skip list, literal / registry-URI / unknown-alias pass-through, a synthetic `services:` block, in-place mutation with per-resolution notes, and no-`containers`-map behaviour. The existing nginx / benchmark / exporter alias tests in `test_configs.py` still pass unchanged.

## Stack

1. #386 remove `--bash`
2. #387 examples matrix
5. #390 `--set` / `--unset`
6. **this PR** one container-alias resolver

### DIFF
--- a/docs/config-reference.md
+++ b/docs/config-reference.md
@@ -118,13 +118,15 @@ The `srtslurm.yaml` file can contain the following fields:
 | `srtctl_root`                   | string | Root directory for srtctl                             |
 | `output_dir`                    | string | Custom output directory (overrides srtctl_root/outputs) |
 | `model_paths`                   | dict   | Model path aliases                                    |
-| `containers`                    | dict   | Container image aliases                               |
+| `containers`                    | dict   | Container image aliases, resolved for every image key in a recipe (see below) |
 | `default_mounts`                | dict   | Cluster-wide container mounts                         |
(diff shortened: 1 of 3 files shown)


ishandhanani:

PR #392, opened 2026-09-06
[2.0] Default resources.gpu_type and gpus_per_node from srtslurm.yaml

ishandhanani:
## Summary

Seventh PR of the 2.0 stack (plan: #385, Track 2 step 9). Stacked on #391; the diff against that branch is what to review.

`gpu_type` and `gpus_per_node` describe the cluster, not the deployment, yet 100% of recipes repeat them. This makes both inheritable so one recipe can move between clusters unchanged.

- `resources.gpu_type` is now optional (`str | None`); a recipe that omits it inherits `srtslurm.yaml` `default_gpu_type`.
- `resources.gpus_per_node` inherits the cluster `gpus_per_node` when omitted (else the existing default of `4`).
- `ClusterConfig` gains `default_gpu_type`.

Applied in `resolve_config_with_defaults`; an explicit recipe value always wins. The fields stay valid in recipes so a run is self-describing for result rollups. No consumer required `gpu_type` to be non-None (it is metadata plus the TRT-LLM numactl `in (...)` check, which handles `None`). Additive: existing recipes are unaffected.

## Validation

- `ruff` clean; full suite: 1763 passed
- `tests/test_configs.py::TestClusterGpuDefaults`: inherit when omitted, recipe value wins, loads with neither recipe nor cluster value
- Cluster: PASSED. srtslurm.yaml got default_gpu_type: b200; job 11924 ran examples/sglang/sglang-router-agg.yaml with resources.gpu_type and gpus_per_node stripped. It submitted (schema accepted the missing gpu_type), and the two aggregated workers launched on CUDA devices 0 and 1, confirming the inherited gpus_per_node=8 drove GPU slicing.

## Stack

...7 of the stack; see #386-#391 for the earlier PRs.

### DIFF
--- a/docs/config-reference.md
+++ b/docs/config-reference.md
@@ -113,7 +113,8 @@ The `srtslurm.yaml` file can contain the following fields:
 | `default_account`               | string | Default SLURM account                                 |
 | `default_partition`             | string | Default SLURM partition                               |
 | `default_time_limit`            | string | Default job time limit                                |
-| `gpus_per_node`                 | int    | Default GPUs per node                                 |
+| `gpus_per_node`                 | int    | Default GPUs per node (applied to recipes that omit `resources.gpus_per_node`) |
+| `default_gpu_type`              | string | Default `resources.gpu_type` for recipes that omit it |
 | `network_interface`             | string | Network interface for NCCL                            |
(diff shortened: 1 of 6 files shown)


ishandhanani:

PR #393, opened 2026-09-06
[2.0] Reject unknown benchmark.type at load; delete dead BenchmarkType enum

ishandhanani:
## Summary

Eighth PR of the 2.0 stack (plan: #385, Track 2 step 4, safe half). Stacked on #392.

An unknown `benchmark.type` (a typo like `gsm8k-bench`, or a removed type) loaded fine and only failed deep in the benchmark stage after a full allocation. `SrtConfig.__post_init__` now validates `benchmark.type` against the runner registry plus the special `manual` type, so a bad type fails at load / dry-run. The import is lazy and guarded, so a registry import hiccup never blocks a load.

The `BenchmarkType` enum listed 10 of the 13 registered types and had zero references; it is removed. The registry (`register_benchmark` / `list_benchmarks`) is the single source of truth.

Scope note: this is the safe half of the benchmark step. It validates the **type**. The per-type field split (rejecting stray fields such as `use_chat_template` on a `custom` benchmark, via a discriminated union) is deferred to a follow-up, because a wrong per-type allowlist would hard-error a valid recipe and needs the golden corpus to land safely.

## Validation

- `ruff` clean; full suite: 1766 passed
- `tests/test_configs.py::TestBenchmarkTypeValidation`: registered + manual types load, `gsm8k-bench` rejected at load, enum is gone

## Stack

8 of the stack; see #386-#392 for the earlier PRs.

### DIFF
--- a/src/srtctl/core/schema.py
+++ b/src/srtctl/core/schema.py
@@ -310,19 +310,6 @@ class Precision(str, Enum):
     BF16 = "bf16"
 
 
-class BenchmarkType(str, Enum):
-    MANUAL = "manual"
-    CUSTOM = "custom"
-    SA_BENCH = "sa-bench"
-    ROUTER = "router"
-    MOONCAKE_ROUTER = "mooncake-router"
-    TRACE_REPLAY = "trace-replay"
-    MMLU = "mmlu"
-    GPQA = "gpqa"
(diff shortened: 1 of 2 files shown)


ishandhanani:

PR #394, opened 2026-09-06
[2.0] Add the roles: authoring surface for recipe topology

ishandhanani:
## Summary

Ninth PR of the 2.0 stack (plan: #385, Track 2 step 5, authoring surface). Stacked on #393.

A recipe can now describe a worker role in one block instead of spreading it across `resources`, `backend.*_environment`, and `backend.<engine>_config.*`:

```yaml
roles:
  prefill:
    nodes: 2
    workers: 6
    gpus: 2
    env:
      PYTHONUNBUFFERED: "1"
    args:
      tensor-parallel-size: 2
      disaggregation-mode: prefill
  decode:
    nodes: 0
    workers: 2
    gpus: 2
    args:
      tensor-parallel-size: 2
      disaggregation-mode: decode
```

- `src/srtctl/core/roles.py::expand_roles` normalizes `roles:` into the existing internal fields (`resources.prefill_workers` / `gpus_per_prefill`, `backend.prefill_environment`, `backend.<engine>_config.prefill`) **before** schema load, wired into `resolve_config_with_defaults` and `SrtConfig.from_yaml` so every load path is covered and no downstream consumer changes.
- Role names are `prefill` / `decode` / `agg`; `agg` maps to the `aggregated` env/config keys; `args` route to the engine's config by `backend.type`; `extra_args` to `<mode>_extra_args` (TRT-LLM).
- `roles_from_legacy` is the inverse, proving the two forms load identically and providing the transform the future migrator will use.
- Mixing `roles:` with the fields it expands into is rejected.

**v1 stays valid.** The legacy layout still loads unchanged, so v1 recipes keep working. `srtctl migrate` still only stamps `schema: 2`; the comment-preserving legacy-to-`roles:` rewrite over the 481-recipe downstream corpus is deferred to the golden-CI migration release, because that structural surgery is where the risk concentrates. Both forms are valid v2.

The 16 topology examples are converted to `roles:`; `features/override.yaml` stays legacy to demonstrate v1 still loads.

## Validation

- `ruff` clean; full suite: 1776 passed
- `tests/test_roles.py` (10): expand produces the legacy layout, roles and legacy load to identical `SrtConfig`, round-trip through `roles_from_legacy`, `agg`→`aggregated`, engine-config key by backend type, TRT-LLM `extra_args`, mixing rejected, unknown role/spec key rejected, no-op without `roles:`, `decode_nodes: 0` survives
- All 17 examples pass `validate_config_file`; the e2e disagg tests pass against the converted examples
- Cluster: PASSED. Job 11936 ran the converted examples/sglang/sglang-router-agg.yaml (roles: agg block). It required fixing preflight to validate the resolved resources (post roles expansion) rather than the raw variant; with that, the recipe submitted and the two aggregated workers launched. The preflight fix has its own test.

## Stack

9 of the stack; see #386-#393 for the earlier PRs.

`env` and `args` are ordinary block-style YAML mappings, written exactly like the legacy `backend.prefill_environment` / `backend.sglang_config.prefill` blocks they normalize into. No inline `{}` or JSON needed; the earlier flow-style snippet here was only shorthand.

### DIFF
--- a/docs/config-reference.md
+++ b/docs/config-reference.md
@@ -197,6 +197,35 @@ model:
 
 ---
 
+## roles
+
+`roles:` is the 2.0 way to describe a worker role. It groups everything about a role in one place instead of spreading it across `resources`, `backend.*_environment`, and `backend.<engine>_config.*`:
+
+```yaml
+roles:
+  prefill:
+    nodes: 2          # -> resources.prefill_nodes
+    workers: 6        # -> resources.prefill_workers
+    gpus: 2           # -> resources.gpus_per_prefill
+    env:              # -> backend.prefill_environment
+      PYTHONUNBUFFERED: "1"
+    args:             # -> backend.<engine>_config.prefill (engine from backend.type)
+      tensor-parallel-size: 2
+      disaggregation-mode: prefill
+  decode:
+    nodes: 0          # 0 shares the prefill node's spare GPUs
+    workers: 2
+    gpus: 2
+    env: { PYTHONUNBUFFERED: "1" }
+    args: { tensor-parallel-size: 2, disaggregation-mode: decode }
+```
+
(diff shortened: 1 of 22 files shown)


ishandhanani:

PR #401, opened 2026-09-06
refactor(observability): retire Hz knobs; collect_interval_ms controls all scrape cadence

nv-yna:
## Summary

srt-slurm's default-on telemetry (tachometer + DCGM/node exporters, on for every run since #358) expressed one concept — *how often measurements are taken* — through several unrelated knobs in mixed units. This PR consolidates them into a single integer knob, **`observability.tachometer.collect_interval_ms`** (milliseconds, default `1000`), retires the Hz-based knobs, and validates with an A/B benchmark on the Dynamo TRT-LLM stack that the shipped defaults add **no measurable performance overhead**, so telemetry can stay on by default.

## The problem

Before this PR, cadence was spread across three unit systems:

| surface | knob | actual unit |
|---|---|---|
| tachometer scrape loop | `observability.tachometer.default_frequency` | **Hz** (scraper slept `1.0/f`) |
| power telemetry collector | `telemetry.default_frequency` | **seconds** — despite the name (passed verbatim into `sample_interval_seconds`; validated against a max-gap-in-seconds constant) |
| DCGM exporter process | `--collect-interval` in its launch command | **milliseconds** |

Two concrete defects followed:

1. **The same key name meant reciprocal things** in adjacent config blocks (`default_frequency: 2` = twice per second under tachometer, once per two seconds under power telemetry).
2. **A 100 ms trap**: a recipe that customized `dcgm_exporter` (even just to pin an image) without spelling out a full `command` silently fell back to the power-telemetry launch template — `--collect-interval=100`, i.e. 10 Hz NVML sampling on every worker node. 10 Hz DCGM sampling was previously measured at **~2% inter-token-latency (p50) overhead** on GB300 decode, which is what motivated unifying and re-measuring cadence in the first place.

## The change

**Commit 1 — retire the Hz knobs.** `collect_interval_ms: int` replaces `default_frequency` on both `observability.tachometer` and `telemetry` (behavior-identical defaults: `1000` = the old 1.0 Hz / 1.0 s). The generated scraper TOML and the vendored Rust scraper switch from per-endpoint `frequency` (Hz, `1/f` sleep) to `collect_interval_ms` (`Duration::from_millis`); the scraper's manual debug flag `--freq` becomes `--collect-interval-ms`. Recipes still carrying `default_frequency` **fail loudly at submit** (`unknown=RAISE`, same retirement pattern as `build_dashboard`/`scrape_metrics`), pinned by tests.

**Commit 2 — one knob drives the whole capture stack.** `observability.tachometer.collect_interval_ms` now governs:

- tachometer's scrape interval for every endpoint (worker, frontend, DCGM, node-exporter pages);
- the launched DCGM exporter's `--collect-interval` — its command is rendered from the knob, so the exporter samples NVML exactly as often as it is scraped: every scrape fresh, no scrape a duplicate. An explicit `dcgm_exporter.command` in the recipe still wins. This kills the 100 ms trap by construction: the power template can no longer leak into a tachometer launch;
- the host sampler's interval (kept — its per-process health output feeds `ingest.py run_host_samples`, which node-exporter families don't cover).

Deliberately **not** coupled: power telemetry's `telemetry.collect_interval_ms` (its dense-sampling fidelity contract inside sa-bench measurement windows is a different purpose, and it must work with tachometer off) and the benchmark client's own server-metrics interval (a separate tool; unrelated to tachometer). Setting the knob below `1000` is honored but warned about at launch (`DCGM_PROVEN_SAFE_INTERVAL_MS`), since faster values drive DCGM sampling toward the measured-harm regime.

## Performance validation (Dynamo TRT-LLM stack)

Four same-day sequential runs on lyris GB300 (DeepSeek-V4 disaggregated, 1 prefill + 3 decode nodes, agentic trace replay, 600 s profiling window, ~200 requests/leg), interleaved **A / F / F / A**: **A** = no telemetry, **F** = this PR at shipped defaults (every cadence at 1000 ms, including DCGM NVML sampling — the previously unmeasured value; 100 ms had measured ~2% overhead, 5000 ms parity). Primary metric: inter-token latency p50, within-arm spread ≤0.4% across all chains of this design.

| metric | A1 | A2 | F1 | F2 | F vs A mean |
|---|---|---|---|---|---|
| **inter-token latency p50 (ms)** | 5.631 | 5.599 | 5.628 | 5.585 | **−0.15%** |
| output tok/s per user | 175.9 | 177.3 | 176.4 | 178.8 | +0.6% |
| output tok/s (total) | 321.7 | 344.2 | 356.2 | 326.4 | +2.5% |
| request throughput (req/s) | 0.284 | 0.317 | 0.357 | 0.306 | +10.3% |
| TTFT p50 (ms) | 758 | 800 | 901 | 551 | −6.9% |
| TTFT p99 (s) | 68.4 | 39.3 | 37.2 | 41.8 | −26.7% |
| inter-token latency p99 (ms) | 7.07 | 7.15 | 7.19 | 6.39 | −4.5% |

**Verdict: parity.** The precision metric is flat (−0.15%); everything else swings within this workload's run-to-run noise (note the largest TTFT p99 outlier sits in a *bare* leg, A1). Zero request errors or cancellations in any leg. Integrity: F legs rendered `dcgm-exporter --collect-interval=1000`, ran tachometer + both exporters (rendered config + process logs); A legs ran zero telemetry processes. Jobs: A1=2946716, F1=2946791, F2=2946891, A2=2947134 (lyris).

Known behavior noted during validation: the scraper sleeps *between* scrape completions, so a slow page stretches its effective period (the DCGM page rendered in ~0.5 s → ~1.5 s effective cadence at a 1000 ms setting). Cosmetic for analysis (timestamps are exact); a fixed-rate ticker is possible follow-up work.

## Testing

- `make check`: 1694 passed, 2 skipped (new: retirement pinning tests for both retired knobs, knob-derivation tests for the DCGM command and host sampler).
- `cargo test --workspace --locked` + `cargo fmt --check` in `src/tachometer/`: all targets green, integration test updated to the new TOML key.
- Mixed-version note: an older `tachometer-scraper` binary reading the new TOML ignores the unknown key and falls back to its built-in 500 ms — degraded but functional; redeploy the binary from this revision to honor configured cadence.

## Migration for recipe authors

| before | after |
|---|---|
| `observability.tachometer.default_frequency: 1.0` (Hz) | `observability.tachometer.collect_interval_ms: 1000` |
| `observability.tachometer.default_frequency: 0.2` (every 5 s) | `observability.tachometer.collect_interval_ms: 5000` |
| `telemetry.default_frequency: 1.0` (seconds) | `telemetry.collect_interval_ms: 1000` |
| custom `dcgm_exporter` block without `command` → silently 100 ms | inherits the knob's cadence; explicit `command` still wins |

### DIFF
--- a/docs/config-reference.md
+++ b/docs/config-reference.md
@@ -1196,7 +1196,7 @@ observability:
 | ---------------- | ---- | ------- | ----------- |
 | `enabled` | bool/null | `null` | `null` means ON for every run (decoupled from `observability.enabled`); explicit `false` opts out |
 | `binary_path` | string | `tachometer-scraper` | Scraper command or path on the compute nodes |
-| `collect_interval_ms` | int | `1000` | Milliseconds between scrapes of every endpoint (replaces the retired Hz-based `default_frequency`) |
+| `collect_interval_ms` | int | `1000` | Milliseconds between scrapes of every endpoint; the single cadence knob — it also drives the launched DCGM exporter's `--collect-interval` (an explicit `dcgm_exporter.command` wins) and the host sampler. Values below `1000` speed up DCGM NVML sampling and are warned about at launch: 100ms sampling measured ~2% decode ITL overhead on GB300. Replaces the retired Hz-based `default_frequency` |
 | `sync_interval_secs` | int | `120` | Interval for intermediate Parquet compaction; `0` disables it |
 | `compaction_threads` | int | `4` | Value passed as `POLARS_MAX_THREADS` |
 | `storage_subdir` | string | `tachometer` | Output directory below the run log directory |
--- a/docs/config-reference.md
+++ b/docs/config-reference.md
@@ -1178,7 +1178,7 @@ observability:
   enabled: true
   tachometer:
     enabled: true
-    default_frequency: 1
+    collect_interval_ms: 1000
     sync_interval_secs: 120
     compaction_threads: 4
     storage_subdir: tachometer
@@ -1196,7 +1196,7 @@ observability:
 | ---------------- | ---- | ------- | ----------- |
 | `enabled` | bool/null | `null` | `null` means ON for every run (decoupled from `observability.enabled`); explicit `false` opts out |
 | `binary_path` | string | `tachometer-scraper` | Scraper command or path on the compute nodes |
-| `default_frequency` | float | `1.0` | Scrape frequency in Hz |
+| `collect_interval_ms` | int | `1000` | Milliseconds between scrapes of every endpoint (replaces the retired Hz-based `default_frequency`) |
 | `sync_interval_secs` | int | `120` | Interval for intermediate Parquet compaction; `0` disables it |
 | `compaction_threads` | int | `4` | Value passed as `POLARS_MAX_THREADS` |
 | `storage_subdir` | string | `tachometer` | Output directory below the run log directory |
@@ -1222,7 +1222,7 @@ When both are enabled, `telemetry.dcgm_exporter` is shared with Tachometer. Do n
 ```yaml
 telemetry:
   enabled: true
-  default_frequency: 1.0
+  collect_interval_ms: 1000
   storage_subdir: power
   required: true
   dcgm_exporter:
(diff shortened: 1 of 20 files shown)


nv-yna:
Second commit (730652ab) per review discussion: **all capture cadence now derives from the single `collect_interval_ms` knob.**

- The tachometer-launched DCGM exporter samples NVML exactly as often as tachometer scrapes it — its `--collect-interval` is rendered from the knob (explicit `dcgm_exporter.command` still wins). This also fixes the trap where an explicit exporter block without a command silently inherited the power template's 100ms — the setting measured at ~2% decode ITL cost.
- The host sampler follows the same knob (kept, not retired: its per-process health data in `host_samples.jsonl` feeds `ingest.py run_host_samples`, which node-exporter families don't cover).
- Values below 1000ms warn at launch (`DCGM_PROVEN_SAFE_INTERVAL_MS`) but are honored.
- Deliberately independent: power telemetry's `telemetry.collect_interval_ms` (own fidelity contract, sa-bench windows) and aiperf's client-side collection interval (separate tool, unrelated to tachometer).

**Merge gate:** DCGM NVML sampling at 1000ms is the one unmeasured cadence (100ms proven harmful, 5000ms proven at parity). A same-day A/F/F/A chain (arm F = this default: exporter at 1000ms) is running on lyris GB300 now; results will be posted here before this PR leaves draft-equivalent state.

PR #403, opened 2026-09-07
Add the readiness probe vocabulary for services: tcp, http, log

ishandhanani:
Stacked on #400 (Track 2 step 8 of https://github.com/NVIDIA/srt-slurm/issues/385). Draft until the stack below it merges.

## What

`services[].readiness` gains the probe vocabulary from the schema design: exactly one of

```yaml
readiness:
  port: 9000                 # shorthand for tcp
```

```yaml
readiness:
  tcp:
    port: 9000               # a TCP connection is accepted
```

```yaml
readiness:
  http:
    port: 8000
    path: /ready             # GET http://<node>:8000/ready
    status: 200              # returns this status
```

```yaml
readiness:
  log:
    pattern: 'Uvicorn running on .*:\d+'   # regex against service_<name>.out
```

plus `timeout_seconds` (default 120) and `interval_seconds` (default 2). The generic wait loop in `core/readiness.py` re-runs the probe until it passes, the deadline expires, or the process dies, so a crashed service fails the job at once with its exit code instead of after the full timeout. The port-collision check now keys on the port a `tcp` or `http` probe implies.

Per the design, the global `health_check` block and the worker and frontend health checks are untouched: this fixes the silent-dead-sidecar class (NVIDIA/InferenceMAX#271) without touching the downstream recipes that set `health_check`.

## Also

- `examples/features/services.yaml` now gates on an `http` probe against the log browser's directory listing.
- `tests/test_readiness.py` covers the schema (shorthand, exactly-one, invalid regex/path/interval), each probe, and the wait loop with a fake clock (retries, fail-fast on death, clipped final sleep). `tests/test_services.py` updated for the probe API.
- Docs: probe section in `docs/services.md`, config-reference row, regenerated schema reference.

## Validation

- Full suite green (1835 passed); lint, schema-docs drift check, every example validates.
- sa-b200, job 12295: the services example gated on `http://<node>:9911/ -> 200` came ready after 40s, the benchmark completed, and the job finished clean (3m25s).

### DIFF
--- a/docs/config-reference.md
+++ b/docs/config-reference.md
@@ -1808,7 +1808,7 @@ services:
 | `env` | dict | `{}` | Service environment; placeholders like `{node_ip}` are substituted |
 | `placement.node` | string | `head` | One instance for `head`/`infra`; one per node for `prefill`/`decode`/`agg`/`workers` |
 | `start` | string | type default | `after_frontend` (generic) or `before_workers` (mooncake-store) |
-| `readiness` | object | none | `port` + `timeout_seconds`; the job waits for it on every service node |
+| `readiness` | object | none | One probe (`port`/`tcp`, `http`, or `log`) plus `timeout_seconds` and `interval_seconds`; the job waits for it on every service node |
 | `inherit_discovery_env` | bool | `true` | Inject the Dynamo discovery env |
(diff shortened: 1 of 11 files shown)


ishandhanani:

PR #404, opened 2026-09-07
Add post_eval: passthrough_env and a command override for the eval dispatch

ishandhanani:
Stacked on #403 (Track 2 step 10 of https://github.com/NVIDIA/srt-slurm/issues/385). Draft until the stack below it merges.

## What

The `RUN_EVAL` / `EVAL_ONLY` evaluation forwards a built-in list of workflow variables into the eval process. Downstream runners extended that list by patching srtctl's source text (`patch_srt_eval_dispatch.py`, anchored on exact lines in `do_sweep.py`), the only runner patch that edits srtctl source. This makes it config:

```yaml
post_eval:
  passthrough_env:          # forwarded into the eval process when set in the job environment
    - EVAL_FRAMEWORK
    - EVAL_CONC
    - EVAL_LIMIT
    - EVAL_SUITE
  command:                  # optional; replaces the built-in lm-eval runner command
    - bash
    - /infmax-workspace/benchmarks/evals/run.sh
    - "{endpoint}"
```

- `passthrough_env` extends the built-in list (`RUN_EVAL`, `EVAL_ONLY`, `MODEL`, `ISL`, `OSL`, `PREFILL_TP`, ...). Names are validated as identifiers.
- `command` replaces the lm-eval runner argv, with `{endpoint}` and `{infmax_workspace}` placeholders, so a runner that rewrites `bench.sh` points at its own script instead of patching. `MODEL_NAME` and `EVAL_CONC` are still set by srtctl.
- `srtctl dry-run` prints the effective dispatch when the block is set.

## Validation

- Full suite green (1841 passed); lint, schema-docs drift check, every example validates.
- `tests/test_post_eval.py` drives `_run_post_eval` with a mocked srun and asserts the forwarded env and the substituted command.

### DIFF
--- a/docs/config-reference.md
+++ b/docs/config-reference.md
@@ -31,6 +31,7 @@ This page is the prose guide: what each block means, how the pieces interact, an
 - [srun_options](#srun_options)
 - [setup_script](#setup_script)
 - [host_setup](#host_setup)
+- [post_eval](#post_eval)
 - [services](#services)
 - [enable_config_dump](#enable_config_dump)
 - [Complete Examples](#complete-examples)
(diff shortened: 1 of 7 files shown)


ishandhanani:

PR #406, opened 2026-09-07
Migrate v1 recipes to the v2 layout and prove equality with srtctl migrate --verify

ishandhanani:
Stacked on #405 (Track 2 step 11 and the deferred full migrator from https://github.com/NVIDIA/srt-slurm/issues/385).

## What

`srtctl migrate` now rewrites the legacy layout, not just the `schema:` key, on a ruamel round-trip document so comments, key order, and quoting survive and moved keys keep their comments:

| v1 | v2 |
| --- | --- |
| `resources.<role>_nodes/_workers`, `gpus_per_<role>`, `backend.<mode>_environment`, `backend.<engine>_config.<mode>`, `backend.<mode>_extra_args` | `roles.<role>.{nodes, workers, gpus, env, args, extra_args}` |
| `frontend.orchestrator_placement` / `dedicated_node`, `benchmark.client_placement` / `client_dedicated_node`, `infra.etcd_nats_dedicated_node` | `placement.node` |
| `dynamo.hash` / `cargo_patches` / `wheel` / `version` | `dynamo.source.{rev, patches, wheel, pypi}` (`top_of_tree` is left with a note; it has no immutable rev) |
| benchmark fields the type never reads | removed (schema 2 rejects them; they were silent no-ops) |

Every variant of an override file (`base`, `override_*`, `zip_override_*`) is rewritten, since a half-migrated file would collide `roles` with the legacy keys on merge. A `backend:` emptied by the folds keeps an explicit `type` instead of disappearing, because a zip variant may `null` its way back to the default and that only works while the base has the key. `-f` accepts directories and repeats; `--in-place` rewrites many files at once.

## Golden equality

`srtctl migrate --verify -f <paths>` migrates in memory, expands plain, override, and sweep files, resolves both documents through the loader, and compares the dumps (masking `dynamo.source`, which only records spelling, and benchmark fields the type never reads). Exit 1 on any mismatch. A new CI job `golden-equality` runs it over `examples/` and the 555 historical recipes extracted from the last commit that carried `recipes/` (`make golden-check` does the same locally).

| Corpus | Identical | Mismatched | Skipped (v1 itself does not load) | Unreadable |
| `examples/` (19) | 19 | 0 | 0 | 0 |
| historical in-repo recipes (555) | 553 | 0 | 2 (`benchmark.type: gsm8k-bench`, unknown on main) | 0 |
| downstream InferenceMAX (481, run locally; private repo) | 370 | 0 | 105 (`telemetry.provider` / `benchmark.tokenizer_mode`, unknown on main) | 6 (duplicate YAML keys, reported by name) |

The skipped and unreadable rows are pre-existing recipe or schema gaps that fail on `main` today, independent of this stack; the report names each one so they can be fixed downstream.

## Also

- Docs: `## schema` and `## roles` in `docs/config-reference.md` describe the rewrite and `--verify`.
- Tests: `tests/test_migrate.py` covers each fold with comments preserved, idempotency, override and zip variants, `dynamo.source` cases, verify outcomes, every example being golden, and the directory forms of `--verify` and `--in-place`.
- Full suite green (1855 passed); lint, schema-docs drift check, every example validates.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -229,3 +229,34 @@ jobs:
               exit(1)
           print(f'\nAll {len(examples)} examples valid')
           "
+
+  # Golden equality: every v1 recipe we know of must resolve to the same config
+  # after `srtctl migrate` rewrites it to schema 2. Corpus: the curated examples
+  # plus the 555 historical recipes from the last commit that carried recipes/.
+  golden-equality:
+    runs-on: ubuntu-latest
+    env:
+      GOLDEN_RECIPES_COMMIT: e6e9d8b9
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v4
+        with:
+          version: "latest"
+
+      - name: Set up Python
+        run: uv python install 3.10
+
+      - name: Install dependencies
+        run: uv sync --dev
+
+      - name: Extract the historical recipe corpus
+        run: |
+          git fetch --depth=1 origin "$GOLDEN_RECIPES_COMMIT"
+          mkdir -p /tmp/golden
+          git archive "$GOLDEN_RECIPES_COMMIT" recipes | tar -x -C /tmp/golden
+          find /tmp/golden/recipes -name '*.yaml' | wc -l
+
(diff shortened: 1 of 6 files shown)


ishandhanani:

PR #411, opened 2026-09-09
fix(profiling): support Dynamo Nsight capture

kyleliang-nv:
## Summary

Fix iteration-triggered Dynamo Nsight capture by using every selected worker's `DYN_SYSTEM_PORT` and the correct `/engine/control/start_profile` and `/engine/control/stop_profile` routes.

- Preserve existing all-worker wrapping: per-phase `capture_scope` defaults to `all`. Targeting one `worker_index` / `worker_rank` requires explicit `capture_scope: selected`.
- Validate selected physical worker/rank bounds before submission.
- Preserve leader-only control for native Dynamo sidecars and direct vLLM: all-process capture wraps followers but sends control once through the leader.
- Prepend optional Nsight library paths inside the container without replacing its `LD_LIBRARY_PATH`.
- Expose non-TRT-LLM Nsight trace/range options and an Ubuntu CLI setup script for x86_64 and Arm64.

## Backend scope

Dynamo-hosted vLLM/SGLang use the system-server control routes. Direct SGLang retains its native `/start_profile` and `/stop_profile`; versions without a `start_step` request field start immediately and honor only the derived `num_steps` (pre-existing behavior).

TRT-LLM remains executor-driven through `TLLM_PROFILE_START_STOP`, without benchmark-side HTTP control. The new `nsys_trace` and `capture_range_end` fields remain explicitly non-TRT-LLM in this PR.

For vLLM `dp_launch_mode: per_gpu`, each physical DP process has its own Nsight wrapper/report. With `per_node`, one wrapped process owns multiple local DP ranks. Targeting selectors are ignored in `capture_scope: all`; validation now warns when either ignored selector is nonzero.

This is a generic Dynamo profiling fix; it adds no benchmark-specific lifecycle changes.

## Review updates

- Rebased onto the schema-v2 main (`1425eeb6`) and regenerated the schema reference; both generated documents pass the freshness check.
- Warn when implicit or explicit `capture_scope: all` ignores a nonzero worker index/rank. Keep all-worker behavior unchanged; selected scope remains explicit. Added 36 cases covering both vLLM/SGLang, all phases, and default/explicit all scope.

- Restored backward-compatible all-worker capture as the default and made selected-process tests explicitly opt in.
- Added real-topology regression coverage: two logical workers spanning four nodes, with both per-GPU and per-node layouts, validating physical ranks and every allocated system port.
- Documented validation-time topology/port-allocation assumptions.
- Removed unrelated signature/docstring formatting churn.
- Documented using `nsys_trace` instead of duplicating `--trace` in `extra_nsys_args`.
- Kept both execution-extension display guards: the outer OR also covers non-profiling extensions; the inner guard is necessary. Tests cover profiling-only display and custom-benchmark display with profiling disabled.

## Relationship to #264

Includes the Dynamo route correction independently proposed in #264, plus physical-worker targeting, all-process control, topology-aware validation, and regression coverage. It does not depend on #264.

## Validation

Rebased on `main` at `1425eeb6`; current revision `807ec40`.

- Focused profiling/schema-docs/dry-run/measurement-window suite: **202 passed**.
- Both generated schema documents pass `srtctl schema-docs --check`.
- All **21 current examples** validate.
- Source Ruff lint/format, profiling-test lint/format, shell syntax, and `git diff --check` pass.
- Full local suite: **2,419 passed, 2 skipped, 6 deselected**. All six local failures reproduced on unchanged `main` at `1425eeb6` in the same environment (macOS CPU-affinity, shell-path, and terminal-formatting assumptions).
- Local `ty` diagnostics are identical to unchanged `main`, apart from shifted line numbers; the CI type-check step remains advisory.
- Fresh [GitHub Linux CI](https://github.com/NVIDIA/srt-slurm/actions/runs/34785981846) is green on `807ec40`: full tests and coverage, lint/schema-doc checks, mock/server integration, examples, migration golden equality, and the advisory type-check job. The separate copyright check also passed.

No new GPU/Nsight hardware capture was performed for this review revision.


src/srtctl/core/schema.py
```diff
@@ -828,6 +828,9 @@ class ProfilingPhaseConfig:
 
     start_step: int | None = None  # Step to start profiling
     stop_step: int | None = None  # Step to stop profiling
+    capture_scope: Literal["selected", "all"] = "selected"
```
zbpatel: Defaulting `capture_scope` to `"selected"` (with `worker_index=0`/`worker_rank=0`) is a silent behavior change for existing `type: nsys` (iteration-triggered) recipes on multi-worker / multi-rank vLLM/SGLang. Before this PR, `worker_stage` wrapped *every* process (`if profiling.is_nsys:`) and the benchmark received *all* leader endpoints via `_logical_worker_endpoints()`. After this PR, an existing recipe with only `start_step`/`stop_step` set will, with no config change, wrap only worker 0 / rank 0 and send only that one control endpoint to the benchmark.

Two options:
- Default `capture_scope: "all"` so existing recipes keep wrapping everything, and opt *into* targeted capture explicitly; or
- Keep `"selected"` but call this out loudly in the PR description / a changelog note with the one-line migration (`capture_scope: all`).

I'd lean toward defaulting to `"all"` for backward compat — targeted capture is the new capability, so it's the one that should opt in. If `"selected"` stays the default, please add a test pinning the default-on-multi-worker behavior so the change is intentional and visible.

src/srtctl/core/schema.py
```diff
@@ -860,6 +863,21 @@ class ProfilingConfig:
     # Extra arguments passed to nsys profile (appended before `-o`; see get_nsys_prefix)
     extra_nsys_args: list[str] | None = None
 
+    # Non-TRT-LLM Nsight activity domains. ``cuda-sw`` can be selected
+    # explicitly where software tracing is preferred over hardware tracing.
+    nsys_trace: str = "cuda,nvtx"
```
zbpatel, line 936: `nsys_trace` and `capture_range_end` are scoped to non-TRT-LLM only — `_get_nsys_prefix_trtllm` still hardcodes `cuda,nvtx,ucx` / `stop` / `--sample=none` / `--cuda-graph-trace=node`, so setting either field on a TRT-LLM run is silently ignored.

To be clear, this isn't a regression: the TRT-LLM hardcoding is preexisting, and `extra_nsys_args` (which *is* applied on the TRT-LLM path) still works as before. It's just that the two new fields don't apply there.

This will be addressed in #412, which routes all of these flags through shared helpers (`_nsys_trace_domain`, `_nsys_cuda_graph_trace_mode`, `_nsys_sample_arg`) across all four `get_nsys_prefix_*` branches, so the override surface is uniform across backends and there's no silent non-TRT-LLM-only scope. No action needed in this PR — flagging so reviewers know the gap is closed in the follow-up.

src/srtctl/core/schema.py
```diff
@@ -900,13 +918,16 @@ def _get_phase_config(self, mode: str) -> ProfilingPhaseConfig | None:
             return self.aggregated
         return None
 
-    def get_env_vars(self, mode: str, profile_dir: str) -> dict[str, str]:
+    def get_env_vars(
```
zbpatel: This hunk rewraps `get_env_vars` from a one-line signature to multi-line and deletes the blank line between the `Args` block and `Returns:` in the docstring. It's unrelated to the PR's goal, adds diff noise, and removing the `Args`/`Returns` blank line breaks Google-style docstring section separation (and may trip a docstring linter). Please revert this hunk.

src/srtctl/core/schema.py
```diff
@@ -2258,6 +2295,32 @@ def _vllm_has_mooncake(mode_cfg: dict | None) -> bool:
                     "for you."
                 )
 
+    def _profiling_worker_ranks(self, mode: Literal["prefill", "decode", "agg"]) -> set[int]:
```
zbpatel, line 2706: Building a synthetic `Endpoint` with fake node names and delegating to the real `backend.endpoints_to_processes` is a neat way to derive the valid rank set without duplicating per-backend topology logic — and it works because `Endpoint.gpu_indices` is per-node and the per_gpu DP branch keys `node_rank` off `dp_rank` driven by the `gpu_indices` count.

Two small asks:
1. Add a one-line comment noting the assumption it relies on — e.g. `# mirrors Endpoint's uniform-per-node gpu_indices assumption; gpus_per_worker must be a multiple of gpus_per_node for multi-node workers` — so the next reader doesn't have to re-derive it.
2. It runs a full `endpoints_to_processes` (including port allocation) during validation just to read `node_rank`. Cheap, and the fake node names isolate any shared port allocator, but worth a brief note that this is validation-time work.

src/srtctl/cli/submit.py
```diff
@@ -432,6 +433,41 @@ def show_config_details(config: SrtConfig) -> None:
         if config.benchmark.container_image:
             details.add_row("benchmark", "container_image", config.benchmark.container_image)
 
+        profiling = config.profiling
+        if profiling.enabled:
```
zbpatel, line 572: `show_extensions` now gates on `or config.profiling.enabled` (line 416), so the inner `if profiling.enabled:` on line 437 is redundant. Drop one of the two checks.

docs/profiling.md
```diff
@@ -84,6 +94,14 @@ profiling:
 | `decode.stop_step`      | Step number to end decode profiling           | `50`     |
 | `aggregated.start_step` | Step number to begin aggregated profiling     | `0`      |
 | `aggregated.stop_step`  | Step number to end aggregated profiling       | `50`     |
+| `*.capture_scope`       | Capture one selected process or all physical processes | `selected` |
+| `*.worker_index`        | Logical worker selected for iteration-based nsys | `0`    |
+| `*.worker_rank`         | Physical process rank selected within that worker | `0`  |
+| `nsys_trace`            | Non-TRT-LLM Nsight activity domains          | `cuda,nvtx` |
```
zbpatel, line 100: `nsys_trace` and `extra_nsys_args: ["--trace=..."]` both target the same `nsys -t` flag. A user setting both (e.g. `nsys_trace: "cuda,nvtx,osrt"` plus `extra_nsys_args: ["--trace=osrt"]`) will pass two `-t`/`--trace` flags and rely on nsys's last-wins behavior. Worth a one-line note here recommending `nsys_trace` over `--trace=` in `extra_nsys_args` when the dedicated field exists.

zbpatel (commented):
Inline review comments from the overlapping profiling work.

src/srtctl/core/schema.py
kyleliang-nv: Fixed in 7ac824c: capture_scope now defaults to all, preserving existing all-worker wrapping. Targeted capture requires explicit selected. Added a regression that runs the real allocator for two logical workers across four nodes, with both per_gpu and per_node layouts, and verifies all physical ranks are wrapped and every allocated DYN_SYSTEM_PORT reaches the helper.

src/srtctl/core/schema.py
kyleliang-nv: Addressed in 7ac824c. The comment now explains that this mirrors allocate_endpoints: multi-node workers get a uniform full per-node GPU set and partial-node workers a contiguous subset. A second comment notes that validation expansion uses a fresh port allocator, without reserving live ports or consuming runtime allocations. Added tests comparing the validation rank set with real allocated per_gpu/per_node topologies and their system ports.

src/srtctl/cli/submit.py
kyleliang-nv: I retained both guards after checking the surrounding condition: show_extensions is an OR of custom benchmark/container, observability, telemetry, Mooncake, and profiling. A custom benchmark can make the outer condition true while profiling is disabled, so removing the inner check would display disabled profiling. Removing profiling from the outer condition would hide profiling-only configurations. Added a clarifying comment and regression assertions for both cases in 7ac824c.

src/srtctl/core/schema.py
kyleliang-nv: Agreed; keeping this PR scoped to Dynamo capture. The schema/docs/PR description explicitly mark nsys_trace and capture_range_end as non-TRT-LLM. No TRT-LLM behavior change here.

docs/profiling.md
kyleliang-nv: Added the note in 7ac824c: for non-TRT-LLM workers prefer nsys_trace and do not also pass --trace through extra_nsys_args; otherwise duplicate options depend on Nsight argument parsing.

src/srtctl/core/schema.py
kyleliang-nv: Fixed in 7ac824c: restored the single-line get_env_vars signature and the blank line between Args and Returns.

PR #412, opened 2026-09-09
feat(profiling): expose common nsys flags

zbpatel:
Extend the Nsight options introduced in #411 across all backends. Explicit `nsys_trace` and `capture_range_end` settings now also apply to TRT-LLM; `cuda_graph_trace_mode` and `sample_mode` configure graph tracing and CPU sampling for both iteration- and time-based capture.

Use the merged `nsys_trace` field instead of introducing a second `trace_domain` setting. Omitted/null trace settings retain `cuda,nvtx,ucx` for TRT-LLM and `cuda,nvtx` elsewhere. Sampling remains `none` on TRT-LLM and omitted elsewhere; graph tracing remains `node`. Capture-range behavior defaults to `stop` and applies only to iteration-based capture.

Update dry-run output, user documentation, and generated schema documentation.

Validation: 215 tests passed across profiling, dry-run, schema documentation, and SA-Bench measurement-window suites. Ruff lint, formatting, schema freshness, and diff checks passed. GPU/Slurm runtime capture was not exercised locally.


zbpatel:
Ok I just updated this to be compatible with @kyleliang-nv 's PR !411 that was just merged in. It should be ready to go now

PR #431, opened 2026-09-13
feat: [AMD Support via SemiAnalysis][1/4] AMD hardware and required cluster runtime

cquil11:
## Scope

Enable AMD hardware and make native srt-slurm jobs work on our Slurm clusters. This combines the former hardware and required-runtime PRs; functionality needed to launch the deployment is no longer hidden in a separate prerequisite layer.

- Cluster configuration owns GPU visibility, GPU-exporter defaults, network interface, mounts, model paths, and Slurm options. NVIDIA defaults are preserved. There is no vendor enum, accelerator dispatch module, or new GRES convention.
- Resolve GPU-exporter defaults once at config load, with explicit recipe overrides taking precedence.
- Apply the configured visibility variable to allocated GPU subsets; preserve full-node and backend device-binding behavior. vLLM has one `set_visible_devices` option.
- Respect the configured interface for control-plane and SGLang worker addresses, and forward recipe Slurm options to native routers.
- Write worker configuration inside the existing container log mount.
- Static routers that need ready HTTP workers opt into the existing health polling before launch. SGLang enables it so slow model loading cannot exhaust initial registration. Worker URLs are resolved once and reused.

Upstream scheduling escape hatches handle GRES: `use_gpus_per_node_directive: false` and `default_sbatch_directives: {gres: "gpu:8"}`. Cluster host setup uses the existing lifecycle. No engine patch, new deployment system, automatic network fallback reordering, or Dynamo installation workaround is added.

## Review map

- `core/schema.py`, `core/config.py`, `core/runtime.py`: cluster defaults and runtime context.
- `cli/mixins/worker_stage.py`, `backends/vllm.py`: GPU binding and container-visible config.
- `frontends/`, `services/implicit.py`: configured addresses, Slurm options, and native startup.
- Example profile, dry-run, docs, and focused behavioral tests accompany those changes.

## Manual stack

**#431 AMD hardware + required cluster runtime → #432 vLLM/MoRI → #433 ATOM/AToMesh → #476 optional observation and diagnostics.**

The core runtime does not depend on #476. The InferenceX CI launcher uses its optional `srtctl wait` command to stream logs and report allocation completion; it is not part of GPU/worker/router enablement. ATOM uses Mooncake and has no engine dependency on MoRI.

Optional deployment [fork #16](https://github.com/SemiAnalysisAI/srt-slurm/pull/16), Dynamo installation [fork #15](https://github.com/SemiAnalysisAI/srt-slurm/pull/15), and Infera [fork #3](https://github.com/SemiAnalysisAI/srt-slurm/pull/3) are outside this validation.

## E2E validation

[InferenceX #2555](https://github.com/SemiAnalysisAI/InferenceX/pull/2555) targets the seven-point MI355X DeepSeek V4 AgentX sweep plus three default evaluations.

Current full validation: [InferenceX run 35231968727](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/35231968727), port commit `15e61d4e865c9eb8c64009f57ec0948e4c934e12`, fork runtime `c63427249ab996bd38d364cad1629aa5aa8877ef`. This preserves the maintainer-selected [DSPARK baseline #3188](https://github.com/SemiAnalysisAI/InferenceX/pull/3188), including all seven points and real-acceptance evaluation. Not yet green.

[Previous attempt 35181888533](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/35181888533) passed four throughput points and c16 GSM8K (97.35% versus 91% threshold). Three points stopped during host preparation, and two evals exposed a JSON-environment templating defect in the optional runtime layer. The replacement removes that formatting, uses native eval artifacts, and restores the existing GPU-drain time allowance. Serving settings are unchanged. #3170 is closed as superseded.

**The reorganized revision is not E2E validated.** These upstream PRs are synchronized for review; a complete green run and artifact inspection are still pending.

Latest completed runs before this cleanup/rebase, retained as historical evidence only:

- [SGLang AgentX full sweep](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/33464520100)
- [SGLang fixed-sequence full sweep](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/33554199275)
- [ATOM/AToMesh full sweep](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/33825304569)
- [vLLM MiniMax M3 validation](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/33837838363)

All four fork PRs target upstream `main` and are manually stacked: **#431 → #432 → #433 → #476**. Each later branch includes the preceding layers; merge in order and rebase the remaining PRs.


src/srtctl/core/accelerator.py
```diff
@@ -0,0 +1,23 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Accelerator-specific runtime environment helpers."""
+
+from typing import Literal
+
+AcceleratorVendor = Literal["nvidia", "amd"]
```
ishandhanani: `Literal["nvidia", "amd"]` is spelled out again in `core/runtime.py:241` and `core/schema.py:302` instead of importing this alias, so adding a vendor means editing three files plus the chain below. If the enum stays, the other two should import `AcceleratorVendor`.

src/srtctl/core/accelerator.py
```diff
@@ -0,0 +1,23 @@
...
+    ROCm recommends ``ROCR_VISIBLE_DEVICES`` for GPU isolation on Linux.  Do
+    not also set ``HIP_VISIBLE_DEVICES`` or ``CUDA_VISIBLE_DEVICES`` here:
+    those variables are interpreted after ROCr device masking by some stacks,
+    so repeating physical indices can accidentally hide the selected devices.
+    """
+    if vendor == "nvidia":
+        return {"CUDA_VISIBLE_DEVICES": device_ids}
+    if vendor == "amd":
+        return {"ROCR_VISIBLE_DEVICES": device_ids}
+    raise ValueError(f"Unsupported accelerator vendor: {vendor}")
```
ishandhanani: This module is a two-row table from vendor to env var name. The existing pattern for a per-cluster string like this is a plain field on `ClusterConfig` (`network_interface` is the closest analogue): `visible_devices_env: str = "CUDA_VISIBLE_DEVICES"`, read once into `RuntimeContext`, with an AMD profile setting `ROCR_VISIBLE_DEVICES`. That removes this file, the `ValueError` path, and the `Literal` duplication, and it does not close the door on a third vendor.

Neither #432 nor #433 adds further vendor branches beyond these two and the DCGM check, so nothing downstream needs the enum yet.

src/srtctl/core/schema.py
```diff
@@ -299,6 +299,7 @@ class ClusterConfig:
     # recipe move between clusters of different GPU types without an edit.
     default_gpu_type: str | None = None
     network_interface: str | None = None
+    accelerator_vendor: Literal["nvidia", "amd"] = "nvidia"
```
ishandhanani: See the comment on `core/accelerator.py`. I would replace this with `visible_devices_env: str = "CUDA_VISIBLE_DEVICES"` and a `default_gpu_exporter: TelemetryExporterConfig | None = None` block next to the other `default_*` fields below. Both are ordinary cluster knobs that follow patterns already in this class, and together they cover everything `accelerator_vendor` does today.

src/srtctl/core/schema.py
```diff
@@ -1221,9 +1222,13 @@ class TachometerConfig:
...
     @property
     def resolved_dcgm_exporter(self) -> TelemetryExporterConfig | None:
-        """User-configured DCGM exporter, else the built-in default."""
+        """User-configured exporter, else the NVIDIA-only built-in default."""
+        from srtctl.core.config import get_srtslurm_setting
+
         if self.dcgm_exporter is not None:
             return self.dcgm_exporter
+        if get_srtslurm_setting("accelerator_vendor", "nvidia") != "nvidia":
+            return None
```
ishandhanani: Three issues with reading the cluster config from inside a schema property:

- It bypasses `runtime.accelerator_vendor`, which this PR adds as the canonical copy, so there are now two sources of truth for the vendor. `RuntimeContext` exists so derived values are computed once.
- `load_cluster_config()` (`core/config.py:55`) is uncached: every call re-opens, parses, and marshmallow-validates `srtslurm.yaml`. This property is evaluated from seven call sites (`core/telemetry.py:79`, `services/implicit.py:96`, `cli/mixins/telemetry_stage.py:551` and `:580`, `cli/submit.py:537`, `core/schema.py:2744`).
- The policy itself is hardcoded. An AMD cluster gets no GPU exporter by default and has no cluster-level way to declare one, so every AMD recipe must set `observability.tachometer.dcgm_exporter` by hand (and the field is still named dcgm). A `default_gpu_exporter` block on `ClusterConfig`, following `default_health_check` / `default_host_setup`, would let an AMD profile name its exporter once, and this property would go back to being pure: recipe field, else cluster default, else `DEFAULT_DCGM_EXPORTER`.

ishandhanani (commented):
The net diff is small (9 files, ~180 lines) and the scope discipline is good: GPU scheduling stays on `use_gpus_per_node_directive` / `default_sbatch_directives`, and the sidecar special case in `worker_stage.py` was only moved, not added. The concern is that the design center, `accelerator_vendor`, is a closed enum whose two effects (pick a visibility env var, suppress the DCGM default) are both hardcoded in Python, when each already has a cluster-config pattern in this repo. srtslurm is meant to be portable across clusters by editing `srtslurm.yaml`, not by adding vendor branches. Inline comments have the specifics.

Suggested shape instead of the enum:

- `visible_devices_env: str = "CUDA_VISIBLE_DEVICES"` on `ClusterConfig`, read once into `RuntimeContext`, same as `network_interface`. An AMD profile sets `ROCR_VISIBLE_DEVICES`. No `accelerator.py`, no `ValueError` path, no `Literal` to keep in sync across three files.
- `default_gpu_exporter: TelemetryExporterConfig | None` on `ClusterConfig`, following `default_health_check` / `default_host_setup`. An AMD profile points it at AMD's device-metrics-exporter once; NVIDIA clusters keep the built-in DCGM default; `observability.tachometer.dcgm_exporter` in a recipe still wins. The schema property then stops reading global config.

Not in the diff, but worth noting for a PR titled base AMD support:

- `src/srtctl/core/fingerprint.py:330` and `:917` shell out to `nvidia-smi`; `:353-363` call `nvcc` and `torch.cuda.nccl.version()`. On AMD the fingerprint and lockfile silently record GPU, CUDA, and NCCL as unavailable.
- `srtslurm.yaml.example` is not updated here (#432 adds the line).
- The `CLAUDE.md` hunk is a trailing-newline removal only.

Housekeeping:

- The first commit touched ~30 files and the second reverted nearly all of them. Please squash so the history matches what lands.
- CI `conventional-title` is red on the `[AMD Support][1/3]` prefix.

src/srtctl/core/runtime.py
```diff
@@ -238,6 +238,7 @@ class RuntimeContext:
     # HuggingFace model support - True if model.path was "hf:model/name"
     is_hf_model: bool = False
     gpu_type: str | None = None
+    accelerator_vendor: Literal["nvidia", "amd"] = "nvidia"
```
ishandhanani: Duplicated `Literal`; import `AcceleratorVendor` from `core/accelerator.py` if the enum stays. Or, per the `ClusterConfig` comment, this becomes `visible_devices_env: str` and the `Literal` goes away.

src/srtctl/backends/vllm.py
```diff
@@ -263,7 +263,11 @@ class VLLMProtocol:
     # vLLM server CLI config per mode
     vllm_config: VLLMServerConfig | None = None
 
-    # Legacy device binding for vLLM builds without --device-ids.
+    # Vendor-neutral device binding for vLLM builds without --device-ids.
+    # When unset, preserve the legacy CUDA-named option below.
+    set_visible_devices: bool | None = None
+
+    # Legacy compatibility alias. New recipes should use set_visible_devices.
     set_cuda_visible_devices: bool = False
```
ishandhanani: Two booleans to express one tri-state is fussy: every reader now has to know about both fields. A single `set_visible_devices: bool | None` with the legacy `set_cuda_visible_devices` folded into it at load time (the way `services/normalize.py` maps legacy keys before schema load) would keep the alias without the dual-field precedence rule. Also `should_set_visible_devices` at line 752 takes `process` and never reads it.

tests/test_accelerator.py
```diff
@@ -0,0 +1,97 @@
...
+
+def test_amd_default_services_omit_nvidia_exporter(monkeypatch) -> None:
+    """Upstream's default-on capture must not start NVIDIA software on AMD."""
+    from srtctl.core.schema import ModelConfig, ResourceConfig, SrtConfig
+    from srtctl.services.implicit import effective_services
+
+    monkeypatch.setattr(
+        "srtctl.core.config.get_srtslurm_setting",
+        lambda key, default=None: "amd" if key == "accelerator_vendor" else default,
+    )
```
ishandhanani: The patch target is coupled to the function-local import inside `resolved_dcgm_exporter`; hoisting that import to module scope breaks this test. If the vendor or exporter default came from a `ClusterConfig` field carried on `RuntimeContext`, the test could construct the config directly without monkeypatching a global.

ishandhanani (approved):
bet

ishandhanani:
Merged `main` into this branch (5b61ec67) so it lands cleanly on top of #486, #487, and #489. Three files conflicted and were resolved by hand; the rest of the branch is unchanged.

What moved, and why:

- **`should_set_visible_devices()` is now a `BackendProtocol` member.** #489 made every backend question go through the protocol (no `getattr(backend, ...)` in `src/`, enforced by `tests/test_backend_protocol.py`). The rename from this PR stays; SGLang, TRT-LLM, and the mocker answer `True`, vLLM answers from `set_visible_devices`.
- **`_visible_device_environment` reads `self.backend.should_set_visible_devices()` and `self.config.dynamo.sidecar` directly.** Same behavior as before, without the `getattr` defaults. Both worker launch paths build the mask from `runtime.visible_devices_env` through this one helper.
- **Static router pre-start probe reads `config.health_check` and passes `runtime.srun_options` directly.** The `MockConfig` test fake gained `health_check: None`, which is how tests opt out of the probe; the probe itself is covered by `test_sglang_router_starts_only_after_workers_are_healthy`.
- **`SGLangRouterFrontend.resolve_worker_host` passes the network interface.** The base class already did after #487; the override now matches.

Validation on the merged head: full suite 3034 passed on Python 3.13 and 3.10, `make lint` scope clean, schema docs in sync, and `srtctl dry-run` on all 26 example recipes differs from `main` only by the `GPU subset visibility variable` line this PR adds.

cquil11:
Passed: [MI355X DeepSeek V4 DSpark smoke](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/35767066251), SGLang + MoRI 1P1D TP4, c4, on fork stack `7a7ca652`. 190 measured requests succeeded, zero request errors; one end-of-window cancellation. Throughput smoke only, no accuracy eval.

PR #432, opened 2026-09-13
feat: [AMD Support via SemiAnalysis][2/4] vLLM MoRI-IO integration

cquil11:
## Summary

Integrate native vLLM Router with MoRI-IO prefill/decode workers. Static router registration and NIXL port handling do not describe MoRI's ZMQ discovery path; this PR supplies the connector-specific launch and readiness behavior.

Stacked on #431, which combines AMD hardware enablement and required cluster runtime. #433 follows this PR; ATOM uses Mooncake and has no engine dependency on MoRI.

## Review map

- `backends/vllm.py`: generate role-aware `MoRIIOConnector` JSON from allocated addresses and ports; preserve upstream ephemeral port allocation for TP children instead of injecting a shared `VLLM_PORT` scan base. Other connectors retain their existing behavior.
- `frontends/vllm_router.py`, `frontends/static_router.py`, `ports.py`: select ZMQ discovery, own its registration port, and omit static P/D URL registration for the managed MoRI path.
- `core/health.py`, `cli/mixins/benchmark_stage.py`: use a one-token completion probe for dynamic discovery, whose registry is not represented by Router's static `/workers` endpoint.
- `core/schema.py`: restrict the managed integration to a P/D layout with one router on the head node.
- Tests and documentation move with the integration. No engine patches, image changes, workload changes, or new GPU tests are introduced by this split.

All four fork PRs target upstream `main` and are manually stacked: **#431 → #432 → #433 → #476**. Each later branch includes the preceding layers; merge in order and rebase the remaining PRs.

[Review only this layer](https://github.com/SemiAnalysisAI/srt-slurm/compare/upstream%2Famd-runtime...upstream%2Fvllm-mori).


ishandhanani:
Merged the updated #431 head into this branch and rebuilt the MoRI-IO layer on what `main` now has (#487 frontend protocol, #489 ports and backend protocol). Same feature, same recipe surface (`engine.connector: moriio`); the implementation moved onto the existing seams. I updated the review map in the description to match. What changed and why, in review-finding order:

1. **Readiness is the Router's own `/health`, not a completions POST.** Since vllm-project/router 43140bc8e2 `VllmPDRouter::health` answers 503 `Waiting for discovered workers` until a prefill and a decode have registered, then 200. `VLLMRouterFrontend.probe_ready` polls it in discovery mode; `probe_ready` now takes the recipe on every frontend so the frontend decides, and `core/health.py` has no `frontend_type == "vllm-router"` branch and no model name to thread through.
2. **Handshake and notify ports come from the allocator.** Upstream defaults every worker to 6301 and 61005, which collide the moment two workers share a node (the repo's own colocated P/D layout). Each discovery worker now gets `moriio_handshake_port` and a `moriio_notify_port` block of one port per local rank (upstream adds the DP and TP rank to the base, `moriio_common.py` 9679173788), plus `host_ip` set to its own routable IP instead of upstream's interface guess. A worker built without these is refused rather than defaulted.
3. **One resolver for the connector.** `connector_for_mode(mode)` covers `engine.connector` and the `roles.<role>.args.connector` override; `uses_moriio()` read only the engine-level field. `discovers_workers()` is built on it, and `validate` requires both roles to agree.
4. **The name lives in the table.** `_CONNECTOR_MAP` rows are `KVConnector` with a `discovery` attribute; `moriio` is a row, and nothing compares the string `"moriio"` in a branch. `kv_transfer_config` builds the JSON from the row and the topology.
5. **Validation moved to the frontend.** The P/D-only, single-router, head-placement rules are in `VLLMRouterFrontend.validate`; the schema only refuses a discovery connector with any other frontend.

Also: `examples/vllm/vllm-router-moriio-disagg.yaml`, a MoRI-IO section in `docs/vllm-router.md` with the contract table, and the duplicate `vllm-router` heading in `config-reference.md` folded into the existing one. Full suite on Python 3.13 and 3.10, and `srtctl dry-run` on all 26 baseline example recipes is unchanged from #431.

One thing I could not verify here: the exact vllm-router version your container ships. Discovery-aware `/health` needs 43140bc8e2 (2026-08-10) or later; with an older router the readiness gate would time out instead of passing early, which is the safe failure.

ishandhanani:
Restacked on #491 (merged) through the updated #431 head. The seams this PR used to carry are now on `main`, so the layer over #431 dropped from 26 files (+762/-74) to 19 files (+614/-28), and its source diff is only the feature: the `moriio` row and its topology JSON in `backends/vllm.py`, the two port kinds and `Process` fields, the Router's discovery command, validation and `/health` probe in `frontends/vllm_router.py`, and the schema guard. Everything else is docs, the example recipe, and tests.

One API note from the restack: `kv_transfer_config(mode, process=None, runtime=None)` keeps #491's one-argument form; only a discovery row reads the topology arguments and refuses to run without them.

Full suite on Python 3.13 and 3.10; `srtctl dry-run` on all 26 baseline example recipes is unchanged from #431. Description updated to match.

ishandhanani:
Merged `main` into this branch (no force-push; your `065da147` is the parent). #431 landed on main as the squash `2917b06a`, so the AMD runtime commit here is now also in main; git kept one copy and the diff against main is exactly this layer again, 19 files. The merge also picks up the TRT-LLM served-model-name forward (#495) and its VRNVL72 NUMA default (#493). `docs/schema-reference.md` was the only conflict and is regenerated.

I checked your two listener fixes against upstream before merging, and both are right:

- **Handshake needs a per-rank block, not one port.** `MoRIIOConnector` computes `side_channel_port = handshake_port + get_port_offset(dp_rank, tp_rank)` (`moriio_connector.py:1664`), so a TP4 worker binds four consecutive ports from its base. My version reserved one, which would have collided between colocated workers at TP>1.
- **26000/27000 instead of 40000/41000.** The old bases sat inside Linux's default ephemeral range, so MoRI's own `bind(0)` listeners (`local_kv_port`, `local_ping_port`) could take a port a later worker's fixed block needs.

Validation on the merge: full suite on Python 3.13 and 3.10, `make lint` scope clean, schema docs in sync, and `srtctl dry-run` over all 26 example recipes byte-identical to main.

One stale line in the description now that #431 is merged: "Stacked on #431" and the "merge in that order" note. Left it to you rather than editing your text.

PR #444, opened 2026-09-14
fix(frontend): use Dynamo's two-tier worker-selection policy for KV routing

ishandhanani:
## Summary

Dynamo ships a second worker-selection policy, `dynamo-two-tier-cost-fn`: it ranks on active-request load first, then KV prefix overlap (the `cache_aware` policy from the experimental SGLang router, ported). Selecting it needs only a `--router-policy-config` YAML, so srtctl now carries that file and passes it by default whenever the Dynamo frontend runs with `args.router-mode: kv`.

- `configs/router-policies/dynamo-two-tier.yaml`: one instance selected for `aggregated`, `prefill` and `decode`; reaches the frontend container through the existing `/configs` mount.
- New `frontend.router_policy` (default `two-tier`). `default` keeps Dynamo's built-in additive cost model; any other value is a container path to your own policy YAML (copy the shipped file to tune `cache_threshold`, `balance_abs_threshold`, `balance_rel_threshold`).
- Not applied when `frontend.args` already set `router-policy-config`, `router-prefill-policy` or `router-decode-policy`, or when the router is not in kv mode.
- **Compatibility:** the shipped policies exist only in ai-dynamo `1.5.0.dev20260908` and newer (no release tag contains them yet), and Dynamo fails at startup if a configured policy type is not linked. When the recipe pins an older PyPI or wheel version (the examples pin `1.4.2`), srtctl logs a warning and leaves the built-in selector in place. Container-provided (`install: false`) and git Dynamo are assumed current.
- `srtctl dry-run` prints `dynamo router policy: two-tier (--router-policy-config ...)`.

## Validation

- New tests in `tests/test_frontends.py`: default in kv mode, nothing for round-robin or other frontends, `default` opt-out, explicit args win, path pass-through, old pinned Dynamo skips with a warning, `start_frontends` appends the flag.
- Full suite green.
- Vera Rubin: Dynamo KV-routed agg job on this branch with the frontend accepting the policy file (job number in a follow-up comment).


ishandhanani:
Vera Rubin, job 87 (`recipes/vr/dynamo-agg.yaml`: Qwen3-0.6B, 2x TP1 `dynamo.sglang`, `--router-mode kv`, image ai-dynamo 1.5.0.dev20260911, `install: false`), this branch at 0f99c314:

- sweep log: `dynamo frontend: --router-policy-config /configs/router-policies/dynamo-two-tier.yaml (frontend.router_policy: two-tier)`
- frontend started and served; Dynamo rejects an unlinked policy type or an invalid policy file at startup, so acceptance means the shipped `dynamo-two-tier-cost-fn` type is linked in that build and the instance resolved
- `Model is ready. Have 0 prefills and 2 decodes.`, benchmark clean: 80 requests, 9104 generated tokens (same as the built-in selector runs of this recipe)

Dynamo logs no INFO line naming the selected worker-selection instance, so the log shows acceptance rather than the policy's routing decisions.

PR #453, opened 2026-09-15
fix(power): one row per CPU socket in the energy report; rail names in one module

FrankD412:
## Summary

**Commit 1 — the bug.** `load_cpu_samples_from` in `power_energy_report.py` keyed the per-socket power series only by `(hostname, socket_id)`. In ACPI mode the collectors write one row **per rail** per socket per instant — Grace/total, CPU rail, SysIO/SoC, DRAM — so all of them landed in the same series as separate "samples" at the same timestamp. `np.trapezoid` then integrated a sawtooth of unrelated rails: same-instant segments have zero width, and the only segment with width was `last_rail(t) → total(t+1)`. DCGM mode emits one row per socket and was unaffected, which is how this slipped through. Fix: group rows by sensor and integrate one rail per socket (total > dcgm > cpu_rail > soc > dram).

**Commit 2 — removing the shape that caused it** (per review). The rail kind was already known at write time in both CPU writers and thrown away, leaving the analysis layer to rebuild it from strings via a third copy of the ACPI label table. Two changes:

1. **Both CPU CSVs pivot to one row per (timestamp, hostname, socket).** `power_w` is the socket's authoritative figure (ACPI `total` envelope or the DCGM value); the ACPI component rails become columns `cpu_rail_w, soc_w, dram_w` (blank for DCGM). Invariant: `sum(power_w over sockets) == total_power_w`. The analysis loader reads `power_w` directly — no rail selection or label knowledge for new files. Scraper `CPU_SCHEMA_VERSION` 1→2, host collector `SAMPLES_SCHEMA_VERSION` 3→4. A socket whose primary rail failed to read is dropped rather than published with a component rail standing in.
2. **`srtctl/core/power/cpu_rails.py` owns the rail vocabulary**: canonical kind names, the single ordered ACPI label→kind pattern table, sensor-name suffixes, legacy exporter `type=` aliases, `classify_sensor()`. `cpu_power_exporter`, `cpu_parser`, the host `AcpiPowerMeterReader`, and `power_energy_report` all import from it. The exporter now publishes canonical `type=total|cpu_rail|soc|dram`; the parser accepts old and new.

Legacy long-format files (every run collected so far) still load: header detection routes them through the commit-1 one-rail-per-socket selection, now sourced from `cpu_rails.legacy_rail_rank`.

DCGM rows deliberately leave the rail columns blank: whether field 1130 corresponds to the ACPI `cpu_rail` or the `total` envelope is unverified (see the `contract.py` NOTE).

## Impact on a real run (job 3054237, 9 Grace nodes, ACPI scraper, 1 h window)

All 18 `cpu_per_socket` rows change; nothing else does. Representative rows:

| socket | joules | avg W | samples | min W | p50 W | max W |
|---|---|---|---|---|---|---|
| theia0200/s0 | 176,677 → **331,030** | 48.8 → **91.5** | 10,856 → **3,620** | 6.0 → 90.6 | 46.8 → 91.1 | 97.0 (unchanged) |
| theia0200/s1 | 198,446 → **375,278** | 54.8 → **103.7** | 10,856 → **3,620** | 5.8 → 101.0 | 53.4 → 102.7 | 117.5 (unchanged) |
| theia0210/s1 | 209,816 → **397,272** | 58.0 → **109.8** | 10,856 → **3,620** | 6.0 → 107.0 | 59.6 → 110.9 | 116.9 (unchanged) |

- samples exactly 3× lower: three rails were being counted as three samples
- min/p5 ≈ 6 W → ≈ 90 W: the old minimum was the SysIO rail leaking in
- max unchanged: the maximum was always a Grace-rail sample, the one stat that happened to be right
- `sum(socket J) == node J` (from `total_power_w`) for every node to the last digit after the fix
- Commit 2 produces a **byte-identical** report to commit 1 on this file; pivoting its 270k v1 rows into the v2 layout gives identical per-socket/per-node series through the wide code path

## Test plan
- [x] `tests/test_cpu_rails.py` (new): label classification incl. Grace `CPU Power` ≠ total, sensor round-trip, legacy OEM labels, legacy `type=` aliases, rank order
- [x] `test_cpu_power_parser`: per-socket pivot with rails; DCGM pivot with no rails; canonical sensor names from OEM labels; legacy `grace/cpu/sysio` aliases; no socket row without a total
- [x] `test_cpu_power`: v4 header pin; `socket_samples` pivot on the ACPI reader; socket dropped when the envelope read fails; collector writes rails/blank rails/total
- [x] `test_cpu_power_samples`: v2 round-trip with rails, exact header/row bytes, v1 legacy read
- [x] `test_cpu_power_collector`: end-to-end via fake HTTP exporter asserts one row per socket and node total
- [x] `test_power_energy_report`: wide layout reads `power_w` directly (mixed ACPI+DCGM nodes); legacy multi-rail rows integrate to the envelope (200 J / 3 samples, not 110 J / 9); component-rail fallback without a total
- [x] `test_cpu_power_contract`: v1 + v2 headers and versions pinned
- [x] 432 tests across power/cpu suites pass; `make lint` unchanged (36 pre-existing, none in touched files)


zhongdaor-nv:
Not sure how much work this is, but it looks like the rail kind is already known at parse time and then dropped before it reaches the CSV:

cpu_parser.py computes it and stores it on the reading: https://github.com/NVIDIA/srt-slurm/blob/34496fc/src/srtctl/core/power/cpu_parser.py#L52-L60
cpu_session.py::_poll builds the CpuSampleRow without it: https://github.com/NVIDIA/srt-slurm/blob/34496fc/src/srtctl/core/power/cpu_session.py#L168-L178
so CPU_SAMPLES_HEADER has no column for it: https://github.com/NVIDIA/srt-slurm/blob/34496fc/src/srtctl/core/power/contract.py#L71-L80

Rather than reconstructing the kind from sensor strings in the analysis layer, could we just add a kind column to the CSV and read it directly?

This would remove the drift risk: the ACPI label knowledge now lives in three places (cpu_power._DOMAIN_PATTERNS, cpu_parser._ACPI_DOMAIN_PATTERNS, and the new _CPU_SENSOR_SUFFIX_KINDS), and a rename in any one of them silently changes every number again and it's same failure mode as the bug being fixed here

FrankD412:
Agreed on both counts — and went one step further in e99f5412. Rather than tag long rows with a `kind` column (which still leaves the reader filtering rows for the same socket), both CPU CSVs now pivot to **one row per (timestamp, hostname, socket)**:

```
…, sensor, socket_id, power_w, cpu_rail_w, soc_w, dram_w, total_power_w, …
```

- `power_w` is the socket's authoritative figure (ACPI `total` envelope, or the DCGM value); `sum(power_w over sockets) == total_power_w`.
- The component rails are columns, populated by ACPI and blank for DCGM (whether field 1130 ≙ `cpu_rail` or the envelope is unverified, so I didn't want to assert it).
- The analysis loader reads `power_w` straight into the socket series. For new files there is no rail selection, ranking, or label knowledge at all — the column *is* the contract, so a rename can't change a number.
- Scraper `CPU_SCHEMA_VERSION` 1→2, host collector `SAMPLES_SCHEMA_VERSION` 3→4. Both readers accept old and new.

On the three-tables drift: new `srtctl/core/power/cpu_rails.py` is the single owner of the kind names, the one ordered ACPI label→kind pattern table, the `CPU<n>:<suffix>` sensor names, and the legacy exporter `type=` aliases. `cpu_power_exporter`, `cpu_parser`, the host `AcpiPowerMeterReader`, and `power_energy_report` all import from it; no other file names a firmware label anymore. (The exporter also now publishes canonical `type=total|cpu_rail|soc|dram` instead of `grace|cpu|sysio`; the parser accepts both.)

Legacy long-format files — i.e. every run collected so far — still load via header detection and the one-rail-per-socket selection from the previous commit, now sourced from `cpu_rails.legacy_rail_rank`. Verified on job 3054237: report byte-identical to the previous commit, and pivoting its 270k v1 rows into the v2 layout gives identical per-socket/per-node series through the wide path.

zhongdaor-nv (approved):
LGTM ship!

peilii (approved):

PR #462, opened 2026-09-16
feat(reporting): S3 upload policy, skip aiperf metric scrapes and compress the rest

ishandhanani:
`reporting.s3` synced the whole log directory. Measured on real runs, a sa-bench job is about 250 MB and the long trace replays 2.1 to 2.2 GB, over 95% of it aiperf's per-interval scrape of the worker and DCGM `/metrics` endpoints (`server_metrics_export.jsonl`), stored twice: raw per concurrency under `artifacts/` and reshaped again inside `perf_dashboard_bundle/`. Tachometer already records that series as parquet at a fraction of the size, the bundle is a re-renderable intermediate, and `perf_dashboard.json` duplicates the self-contained HTML. The same directory compresses about 60:1. This PR gives the upload a policy: skip what is duplicated elsewhere, compress the one raw file worth keeping, ship the rest as before.

### How This Was Implemented
- `S3Config.exclude` (AWS `--exclude` patterns, relative to the log directory, `*` crosses directories) and `S3Config.archive` (Python globs with `**`). Both default to constants in `core/schema.py` (`DEFAULT_S3_EXCLUDE`, `DEFAULT_S3_ARCHIVE`), both are replaced wholesale when set, `[]` disables either.
- Defaults: exclude `server_metrics_export.jsonl`, `server_metrics_export.json`, `gpu_telemetry_export.jsonl` and `inputs.json` under `artifacts/*/` and `sa-bench_*/*/` (the aiperf artifact roots, so a same-named file from another benchmark type is never dropped; scoped after review), plus `perf_dashboard_bundle/*` and `perf_dashboard.json`; archive `artifacts/**/profile_export.jsonl` and `sa-bench_*/**/profile_export.jsonl` (aiperf's per-request records, the raw truth behind every latency number, 13 to 40 MB raw and under 1 MB compressed). Everything else (run identity, every log, results JSON, `perf_dashboard.html`, tachometer parquet) ships as loose objects as before.
- `postprocess_stage.py`: the container script now installs `awscli` (required) and `zstandard` (best effort), writes the policy into `postprocess-status.json` next to `s3_url`, runs `ARCHIVE_SCRIPT` (an inline stdlib program: glob, tar, zstd or xz fallback, arcnames relative to the log directory, archive written under `/tmp` so nothing is added to the cluster log directory), syncs `/logs` minus the exclude list minus the archived files (`**` collapsed to the AWS wildcard), then `aws s3 cp` the archive beside them.
- Docs: `config-reference.md` gains a `reporting.s3` section with the policy table and the one caveat (with tachometer disabled, dropping the aiperf scrape leaves no engine-metrics record outside the dashboard HTML); schema reference regenerated; CLAUDE.md pointer.

### Validation
- `tests/test_postprocess.py`: schema defaults and loading; the generated bash carries every default exclude as a quoted `--exclude`, the archive patterns as AWS wildcards, the endpoint flag and the policy JSON, and passes `bash -n`; both lists empty reproduces the old whole-directory sync; the archive builder runs for real (`python - <root> <out> <json>`) against a temp tree and packs exactly the matching files with relative names while leaving the copy under `perf_dashboard_bundle/` out, and exits 0 with no output when nothing matches. 77 tests in the file.
- Full suite on Python 3.10: 2538 passed, 2 skipped. `ruff check`, `ruff format --check`, `srtctl schema-docs --check` clean; the two `ty` diagnostics in `schema.py` are pre-existing and unrelated (lines 2591 and 2609).
- Sizing evidence from a real cluster: a 253 MB run tars to 4 MB with zstd level 3; 200 MB of the metrics jsonl becomes 1.6 MB.
- Not run: an end-to-end upload against a live bucket. The `aws s3 sync` and `aws s3 cp` invocations are unchanged apart from the flags, and the archive step was exercised with the same interpreter the container uses.


src/srtctl/core/schema.py
```diff
@@ -191,6 +191,30 @@ def get_prompt(self, log_dir: str) -> str:
...
+# intermediate and holds a reshaped copy of that scrape; ``perf_dashboard.json`` duplicates
+# the self-contained ``perf_dashboard.html``. A 2.2 GB run becomes about 60 MB.
+DEFAULT_S3_EXCLUDE: tuple[str, ...] = (
+    "*/server_metrics_export.jsonl",
+    "*/server_metrics_export.json",
+    "*/gpu_telemetry_export.jsonl",
+    "*/inputs.json",
+    "perf_dashboard_bundle/*",
+    "perf_dashboard.json",
+)
```
KaunilD, line 216: perfect.

fsaady, line 216: minor nit: it might be safer to limit this to aiperf runners, to avoid excluding files with similar names in other modes

PR #463, opened 2026-09-16
feat(telemetry): services declare where their metrics are; tachometer scrapes them there

ishandhanani:
## The idea

Like Kubernetes' `prometheus.io/port` and `prometheus.io/path` annotations: a service says where it serves metrics, and tachometer builds its scrape targets from those declarations. Nothing in tachometer's config generator knows about specific services any more.

```yaml
services:
  - name: envsrv
    type: generic
    command: ["python3", "-m", "server"]
    placement:
      pool: train
    metrics:
      port: 8003          # tachometer scrapes http://<node>:8003/metrics on every node this runs on
```

- **`services[].metrics`** is the annotation: one `{port, path, nodes, name}` or a list of them (`path` defaults to `/metrics`, `nodes` to `all`; `nodes: first` scrapes only the service's first node; `name` becomes the endpoint's name in the parquet and is required when a service declares several, like a ServiceMonitor's `endpoints`).
- **The telemetry stage** turns every endpoint into one target per node the service runs on, pools included: endpoint `<name>_<node>` (the name defaults to the service), row metadata `service=<name>`.
- **Kinds that always publish** return their default from `ServiceKind.metrics()` and set `metrics_filter`, `metrics_endpoint_prefix`, `metrics_gpu_metadata`. The dcgm, node and process exporters are ordinary users of this now: `generate_tachometer_config` loses its exporter special cases and the `exporter_nodes` / `dcgm_exporter` plumbing (about 100 lines). Endpoint names and filters are unchanged, so the parquet schema is the same. Workers and the frontend are not services and keep their own target logic.
- **One special case remains**: power telemetry owning its own DCGM exporter. The stage emits those targets for the worker nodes, and `implicit.py` now implies tachometer's `dcgm-exporter` service unless power telemetry brings one; CPU-only power telemetry used to lose DCGM scraping.
- `srtctl dry-run` prints `metrics=<name>:<port><path>[@first]` per endpoint on every scraped service.

## Why

A Miles run (sa-b200 job 15591) had no SGLang series in tachometer: Miles launches its engines inside the Ray job on ports it picks at runtime, so nothing could name them as targets. The first version of this PR added a dedicated service kind and a runtime script for that; it was too specific. This version makes the general mechanism the change, and Miles a consumer that needs no code.

## Miles, as annotations only

Miles picks its SGLang engines' ports at runtime, so they cannot be named as targets. But Miles's default router, the sgl-model-gateway, serves `GET /engine_metrics`: every engine's samples, each labeled `worker_addr`, on a port the recipe pins with `--sglang-router-port`. Together with Miles's training collector (`--use-prometheus`, a Ray actor pinned to the head, `miles_metric_*` on 9090), the whole integration is a `metrics` list on the `ray` service:

```yaml
services:
  - name: train
    type: ray
    nodes: 2
    metrics:
      - { name: miles, port: 9090, nodes: first }
      - { name: engines, port: 31000, path: /engine_metrics, nodes: first }
```

The Miles example (`examples/miles/qwen3-4b-grpo.yaml`) carries that plus `--sglang-router-port 31000 --use-prometheus --prometheus-port 9090` in the launcher args, with a comment on every knob; `docs/miles.md` has an "Engine metrics" section.

Live with exactly this recipe shape (sa-b200 job 15914, one node, four TP2 engines, three GRPO steps, 14:29): the config carried `miles_im-b200-c013` and `engines_im-b200-c013` only; the parquet holds 1.20 M rows from the router aggregate across 71 `sglang_*` families, each sample labeled `worker_addr=http://172.16.4.105:<port>` for the four engines (548 samples per engine), and 49 k rows from the collector across 83 `miles_metric_*` families, next to 193 k DCGM and 523 k node-exporter rows. Per engine: `sglang_token_usage` p90 0.12 to 0.17, max 0.18 to 0.23; `sglang_num_running_reqs` max 56 to 96; `sglang_gen_throughput` max 18.2 k to 26.5 k tok/s. `miles_metric_rollout_raw_reward` last 0.40, max 0.52; `perf_actor_train_time` 23 to 41 s; `perf_rollout_time` 48 to 49 s. Rollout rewards, KV pressure per engine and GPU power are on one timeline.

## Validation

- Full suite 2664 passed on 3.13 and on 3.10 (lint and schema-docs check too).
- `tests/test_service_metrics.py`: the annotation (including the `nodes` and `name` rules), the exporter kinds' defaults, one target per node across a pool, `nodes: first`, several endpoints on one service, external and disabled services skipped, power-owned DCGM, the generated TOML, dry-run.
- Live on sa-b200: job 15881 (collector alone, `train_<head>` target, 83 `miles_metric_*` families) and job 15914 (collector plus router aggregate, numbers above). Earlier revisions reached the engines through a re-exporting script (jobs 15632, 15640, 15810); that code is gone, the router aggregate covers it.


examples/miles/qwen3-4b-grpo.yaml
```diff
@@ -43,6 +43,15 @@ services:
     type: ray               # one Ray cluster across the job's nodes
     nodes: 2                # this service owns the job's two nodes and sets the allocation size
     preamble: "cd /root/miles"   # the raylet's cwd is what Ray-spawned processes inherit
+  - name: engines               # Miles picks its SGLang engines' ports at runtime; this rider finds them
```
ishandhanani: can we name this engine-metrics maybe?

PR #473, opened 2026-09-17
fix(sglang): relay multinode DP sidecar KV events

jain-ria:
Multinode SGLang attention-DP deployments previously launched a Dynamo sidecar only on the leader, leaving follower KV events unavailable to the router. Launch native gRPC and a coupled sidecar on each follower that owns a KV publisher, while retaining one serving registration per logical worker. Dynamo automatically selects serving or telemetry mode from local engine metadata; srtctl no longer passes `--telemetry-only`. Followers without publishers remain engine-only.

Reserve the full global-DP publisher-port range to prevent collisions between colocated workers, and keep the prefill bootstrap address leader-only. Add two-node aggregated and four-node disaggregated TP2/DP2 examples, documentation, and launch/allocation regression coverage.

Requires [Dynamo PR14908](https://github.com/ai-dynamo/dynamo/pull/14908) and [SGLang PR39659](https://github.com/sgl-project/sglang/pull/39659), including compatible native runtime/gRPC extensions and Python packages. Python-only source overlays are insufficient; the examples require a prepared image.

## GPU validation

Ran both deployments through actual `srtctl apply` on Computelab, using Qwen/Qwen3-0.6B BF16 and one physical A10 GPU per node. The generated SLURM scripts ran unmodified. Normal SA-Bench execution was followed by targeted correctness probes through `post_eval`.

| Deployment | Topology | SLURM job | SA-Bench | Targeted requests |
| Aggregated | One TP2/attention-DP2 worker across two nodes | `4395916` | 40/40, zero errors | 27/27 passed |
| Disaggregated | One TP2/attention-DP2 prefill worker and one TP2/attention-DP2 decode worker across four nodes | `4395931` | 40/40, zero errors | 31/31 passed |

Both jobs completed with exit code `0:0`. SA-Bench used ISL 128, OSL 64, and concurrency 4. Both deployments used NATS request/event planes and Dynamo KV routing; disagg used NIXL/UCX transfer.

The targeted probes and independent checks of saved logs verified:

- Real GPU execution on both DP ranks, on distinct physical nodes; disagg exercised all four prefill/decode nodes.
- KV Stored events from both publisher ranks, distinct publisher IDs under the correct shared serving-worker identity, and events applied by the frontend router. There was one serving registration for agg and one per role for disagg.
- **20/20 unpinned cache-affinity requests per deployment**, across six warmed prompt families, including two streaming requests. Correlated routing logs showed the expected cache-owning rank, positive cache overlap, and success without retries.
- **All four disagg prefill-to-decode rank combinations:** `0→0`, `0→1`, `1→0`, and `1→1`. All benchmark and targeted responses contained generated text and positive completion-token counts.

Disagg enabled KV events on prefill. Decode retained its default disabled prefix cache, so its follower was engine-only; these tests establish prefill cache affinity and transfer to both decode ranks, not decode prefix-cache affinity.

## Revisions and additional checks

- GPU-tested srt-slurm changes were committed as `978b35b`, using Dynamo `9ab50011f8bf` and SGLang `8eaffcf88dac`. Dynamo's native extension was freshly rebuilt; imported native artifacts and source hashes were verified on all four nodes. The SGLang native extension was reused from `c50b251`, whose runtime source is identical to `8eaffcf` (only Python tests changed).
- Native builds used the debug profile with `RUST_MIN_STACK=33554432` (32 MiB Rust thread stacks). This was functional smoke validation; release-build performance and default-stack behavior were not tested.
- After merging main in `6a5497b`, focused regression suites passed **408 tests, 2 skipped**, including SGLang sidecar, allocation, disagg recipe, and vLLM failover coverage. Ruff lint/format and diff checks passed. The merge preserved both SGLang DP port reservations and main's shadow-engine allocation. GPU jobs were not rerun after this merge.
- Dynamo sidecar native unit tests: **54 passed** at `9ab50011f8bf`.


examples/README.md
```diff
@@ -48,7 +52,7 @@ containers:
   dynamo-vllm: /path/to/vllm-runtime.sqsh   # Dynamo vLLM runtime image (ships ai-dynamo and gpu_memory_service), for features/vllm-failover.yaml
 ```
 
-`resources.gpu_type` and `gpus_per_node` are set to `h100` and `8`; change them to match the partition you submit to.
+The matrix recipes set `resources.gpu_type` and `gpus_per_node` to `h100` and `8`; change them to match the partition you submit to. The multinode SGLang sidecar examples allocate one GPU per node: two nodes for aggregate mode, four for disaggregated mode.
```
alec-flowers, line 55: Don't think need this extra comment

PR #477, opened 2026-09-18
feat(cli): add srtctl render for launchers that own the sbatch call

peilii:
## Summary

`srtctl render -f RECIPE --to DIR [-o OUT] [--serve-only] [--setup-script] [--no-preflight]` writes the exact sbatch script `srtctl apply` would submit, without submitting it — for launchers that must own the `sbatch` call themselves (a harness whose contract is `exec sbatch --parsable ...`, reading the job id off that line). The directory receives:

- `sbatch_script.sh` — self-contained: the job copies its recipe from `--to DIR` into `OUTPUT_DIR` at start (`apply` did that copy after sbatch returned; nothing does for a rendered script).
- `config.yaml` (+ `config_<variant>.yaml` for an override variant), git-state snapshot of mounted checkouts.
- `render.json` — what the launcher needs before the job exists: `total_nodes`, `frontend_node_index`, `client_node_index` (computed with the orchestrator's own carving rules via the new `Nodes.planned_role_indices`, parity-tested against `Nodes.from_slurm`), `frontend_port`, `served_model_name`, `benchmark_type`.

Prose goes to stderr; the script path is the last stdout line, so `sbatch --parsable "$(srtctl render ...)"` composes. Sweeps, directories and unselected override files are refused.

Also in this change:

- **`server_ready.json`** is written in `<log_dir>` when every configured worker passes the health gate. An external client driving a `manual` job otherwise only has the frontend to ask, and a Dynamo frontend lists the model as soon as its *first* worker registers.
- **Telemetry samples engine and pool nodes** (`_telemetry_nodes()`: backend-process nodes ∪ `runtime.nodes.compute`). A services-only job sampled nothing and every leg reported itself failed (`expected_nodes: []`, `cpu_collector_failed`).
- **`extra_mount` collisions are warned about**: sources are resolved, so two entries can collapse onto one (`/lustre` → `/scratch` on some clusters) and the later one silently replaced the earlier container path.
- **Telemetry accepts `benchmark.type: manual`** — like serve-only it has no load window, so the whole serve session is captured.

Docs: `docs/cli.md` (render section, TOC), `docs/config-reference.md` (telemetry).

## Validation

- `make lint` clean (ruff check/format); `ty` unchanged.
- `tests/test_render.py` (new, 26 cases): script + staging line ordering, override variant staging, `render.json` contents, 16-way `planned_role_indices` ↔ `from_slurm` parity, CLI refusals (sweep, directory, unselected override), stdout contract, same-file guard, ready marker, `extra_mount` collision warning, telemetry-on-manual. `tests/test_telemetry.py`: node-selection tests; the "manual is rejected" case repointed at `mmlu`.
- Full suite: no new failures relative to `main`.
- On a Slurm cluster (GB300 nodes): rendered scripts submitted with `sbatch --parsable` bring up a 7-node Dynamo/TensorRT-LLM disaggregated deployment and a 1-node services-only job; staged recipe found by the orchestrator, `server_ready.json` written at the gate, `render.json` indices match the allocation, CPU-power telemetry now complete on the services-only job (was `cpu_samples_empty`).


ishandhanani:
doesn't dry-run cover this?

peilii:
Not quite. Three gaps:

1. **dry-run writes nothing.** It renders the script into a rich panel and returns (`submit_with_orchestrator`, the `if dry_run:` block). There's no file for a launcher to hand to `sbatch`.
2. **The script it shows can't be submitted as-is.** `apply` copies the recipe to `outputs/<job_id>/config.yaml` *after* `sbatch` returns the id, and `do_sweep` reads it from there. Submit that script yourself and the job dies at startup with no config. `render` makes the script stage its own recipe from `--to DIR`.
3. **No machine-readable contract.** `render.json` gives the launcher node placement, frontend port and served model name, so it doesn't have to re-derive srtctl's carving rules.

Why a launcher needs this at all: AA's AgentPerf hardware pipeline (ArtificialAnalysis/agentperf-client#151) requires the launch script's last command to be literally `exec sbatch --parsable …` and reads the job id from that line. `srtctl apply` runs sbatch itself and prints a banner, so it can't be used there. Before this we were calling `generate_minimal_sbatch_script` from Python and `sed`-patching the config copy into the script; `render` makes that a supported command.

If you'd rather not add a subcommand, this could be `dry-run --out DIR` instead — the staging line and `render.json` would be identical. Your call.

PR #480, opened 2026-09-18
fix(power): isolate endpoint sampling overruns

Ankur-singh:
# fix(power): isolate endpoint sampling overruns

## Problem

Power collection currently waits for the slowest endpoint before starting the next cycle. One delayed node can therefore create gaps for every healthy node, and the fixed 3-second validation threshold can reject otherwise reliable long-window data. This is causing power validation failures in [InferenceX PR #3087](https://github.com/SemiAnalysisAI/InferenceX/pull/3087).

## Fix

- Poll each endpoint independently on a fixed schedule, with at most one in-flight request per endpoint.
- Record failed and skipped sample slots in compact manifest ranges.
- Derive the normal coverage budget from the configured cadence and request timeout.
- Keep the 10-second absolute and 5% cumulative limits so substantial telemetry loss still fails.

## Why this belongs in srt-slurm

srt-slurm owns the collector, the artifact contract, and the `publication_valid` verdict. Fixing only InferenceX would leave the bad sampling behavior in place, create a consumer-specific exception, and leave every other consumer with the same invalid package.

## Validation

- 520 power and telemetry tests pass.
- Focused type checks, Ruff, and schema documentation checks pass.
- All eight gap-only failures inspected from InferenceX #3087 remain valid under the bounded policy.
- Full suite: 2,907 passed; the same 14 macOS-only failures reproduce on current `origin/main`.


FrankD412:
Thanks for this — the per-endpoint threads are the right fix for the cascade, and making `scrape_seq` a shared slot index (same seed, same arithmetic on every thread, no coordination) is a clean way to keep hosts comparable. I pulled the branch into a worktree, ran the four touched suites (346 pass), then probed the collector with a fake exporter to see how the skip rule behaves at the margins. Two things I'd like to sort out before merge, a few smaller ones after.

## 1. The skip rule forfeits a slot on *any* overrun, including 1 ms

[`_run_endpoint`, session.py L486-487](https://github.com/NVIDIA/srt-slurm/blob/7a87e4dcb13f7ab62f01c7e4a1c121229c47885f/src/srtctl/core/power/session.py#L486-L487):

```python
if not self._stop.is_set() and next_cycle <= now:
    missed_count = int((now - next_cycle) / interval) + 1
```

A slot that came due while the previous request was still in flight is thrown away, even if it's only just due. Measured with a fake endpoint answering in 1.2× the interval (0.06 s on a 0.05 s grid, 1 s run):

| | rows on the slow host | pattern |
|---|---|---|
| main (lock-step) | 5 of 6 slots | cadence stretches to 1.2 s, nobody skips |
| this PR | **12 of 23** slots (52 %) | `ok, missed, ok, missed, …` |
| late-fire variant | **17 of 23** slots (74 %) | `ok, late, late, missed, …` |

So for the case in the title — a node that's a bit slow — the PR halves that node's sample density relative to main. Isolation of the *healthy* nodes works as intended; it's the slow node that pays more than it needs to. The ceiling for a 1.2× endpoint with one in-flight request is 1/1.2 = 83 %, so 52 % leaves a lot on the table.

The four panels below are a deterministic simulation of both rules on a 1 s grid (bars = requests in flight, tick inside a bar = the row's timestamp, dashed boxes = forfeited slots). B1/B2 are the 1.2 s node; C1/C2 are the hang-across-stop case from point 2.

scrape_seq timing: PR #480 skip rule vs late-fire variant, and bracket poll at stop

<sup>Generator and PNG: [`review-assets/pr-480`](https://github.com/FrankD412/srt-slurm/tree/880676266aa2ece67932af5b5c54d47669321cd7/pr-480).</sup>

Proposed: fire the due slot late, forfeit only slots whose *entire* span elapsed.

```python
if not self._stop.is_set() and next_cycle + interval <= now:
    missed_count = int((now - next_cycle) / interval)
```

The top-of-loop `_stop.wait(max(0, next_cycle - now))` already returns immediately when late, so nothing else moves. `scrape_seq` stays a slot index (each seq fires at most once, inside `[N·i, (N+1)·i)`); same-seq skew across hosts is bounded by `interval + Δlatency/2` and re-syncs on every skip — I measured a sawtooth with max 0.077 s on a 0.05 s grid, no drift. Nothing offline joins rows across hosts by seq (coverage validation is timestamp-only), so the in-slot slack doesn't reach the verdict.

Side effect worth noting: with the `+1`, a 2×timeout hang followed by resumption lands the row gap at `interval + 2·timeout + interval` — one interval past the validator's own [`normal_gap_budget = interval + 2·timeout`](https://github.com/NVIDIA/srt-slurm/blob/7a87e4dcb13f7ab62f01c7e4a1c121229c47885f/src/srtctl/core/power/windows.py#L410). With the change the two agree exactly, and the default-config hang case passes on the budget rather than needing the 0.5 %-of-window rule.

This is on my fork as a commit on top of your head, with two tests that fail on the current rule and pass on the new one: [FrankD412/srt-slurm@00174e3d](https://github.com/FrankD412/srt-slurm/commit/00174e3d) (branch [`pr-480-option-a`](https://github.com/FrankD412/srt-slurm/tree/pr-480-option-a), diff vs your head: [`7a87e4dc...pr-480-option-a`](https://github.com/NVIDIA/srt-slurm/compare/7a87e4dcb13f7ab62f01c7e4a1c121229c47885f...FrankD412:srt-slurm:pr-480-option-a)). Cherry-pick it if you agree, or tell me what you'd rather do differently.

## 2. Bracket poll after `_stop` — is a clock-derived seq the right shape?

The in-loop skip block is guarded by `not self._stop.is_set()`. If an endpoint is mid-hang when `stop_and_finalize()` fires, its thread leaves the loop with `scrape_seq` still at `K+1`, never catches up, and does the [bracket poll at `K+1`](https://github.com/NVIDIA/srt-slurm/blob/7a87e4dcb13f7ab62f01c7e4a1c121229c47885f/src/srtctl/core/power/session.py#L502-L507) with `scheduled_at_unix=time.time()`. An idle thread brackets at `S+1` where `S` is the slot containing the stop instant.

With the PR body's own scenario (seq 1 hangs 4 s, stop at 4.5 s): node-a brackets at seq 5, node-b at seq **2**, both at t ≈ 5 s — the one place seq stops being a slot index — and node-b's slots 2–4 never enter `missed_sample_ranges`. `missed_sample_count` under-reports by 3, and `_check_missed_sample_ranges` can't see it since it only checks the ranges that exist.

I want to check the shape with you before writing anything, since it touches the bracket contract. What I have in mind: compute one bracket seq from the clock in `stop_and_finalize` — `B = initial + floor((t_stop − started_monotonic)/interval) + 1` — and have each thread, after leaving the loop, record `scrape_seq…B−1` as `sample_schedule_overrun` if `B > scrape_seq`, then poll `B` with `scheduled_at_unix` = grid time of `B`. Every thread's counter is `≤ S+1` at stop and an idle thread holds exactly `S+1`, so `B` never collides with an existing row and never goes backwards; the `not _stop.is_set()` guard on the in-loop block becomes redundant. Does that match how you're thinking about the bracket, or is there a reason to keep it counter-based?

## 3. `missed_sample_ranges` isn't compact for the steady-slightly-slow case

Coalescing is adjacent-only ([`_record_missed_sample_range`, L401](https://github.com/NVIDIA/srt-slurm/blob/7a87e4dcb13f7ab62f01c7e4a1c121229c47885f/src/srtctl/core/power/session.py#L401)). A persistently 1.2× node alternates hit/miss, so every miss is its own entry: my 1 s probe produced 11; at 1 s cadence over a 2 h window that's ~3,600 entries (~700 KB of manifest). Late-fire drops the miss rate to ~1 in 6 but the shape is the same. Coalescing same-reason ranges within a tolerance, or capping the list while keeping the count exact, would both work.

## 4. `collect_once()` draws from a counter the threads don't advance

[`collect_once`, L291](https://github.com/NVIDIA/srt-slurm/blob/7a87e4dcb13f7ab62f01c7e4a1c121229c47885f/src/srtctl/core/power/session.py#L291) reads `self._scrape_seq`, which only it increments. Called after `_start_collector()` it re-issues seq 0 for every host → duplicate `(seq, host, gpu)` keys → `DUPLICATE_SAMPLE_ROW`. No production caller, but it's public and 18 of the collector tests go through it vs 15 through `start_and_wait_for_readiness`, so over half the suite exercises a path production no longer runs. Suggest it raise if threads are running, or be marked test-only.

## 5. The validation policy change is separable from the threading fix

`normal_gap_budget` is 5 s at defaults, so a 5 s gap now always passes where 3 s was the ceiling on main; on a 30-min window a 9 s gap passes. That's a defensible policy, but it's a different decision from "stop the cascade" and it changes what `publication_valid` means for every consumer. Worth stating explicitly in the PR body — and the eight InferenceX #3087 failures would land better with their actual gap values listed, so a reader can see the 10 s / 0.5 % / 5 % knobs weren't fit to the data.

## Smaller

- `schema.py:3063-3064` (not in this diff) still says `collect_interval_ms … exceeds the 3.0s max sample gap the power validator accepts` — no longer what the validator does.
- [`MAX_SAMPLE_GAP_SECONDS`](https://github.com/NVIDIA/srt-slurm/blob/7a87e4dcb13f7ab62f01c7e4a1c121229c47885f/src/srtctl/core/power/contract.py#L111) is now a config ceiling only; the comment says so, the name still reads like a validator constant.
- Codecov's 43 uncovered lines are almost all the malformed-manifest branches of `_check_missed_sample_ranges`; a handful of bad-manifest fixtures would cover them.
- `scrape_count` changed meaning from "polls performed" to "slots elapsed" (`max(seq)+1`). Nothing contradicts it, but a line in `docs/power-telemetry.md` would help anyone computing `expected_rows = scrape_count × devices`.
- Readiness now requires a shared *slot* rather than a shared *cycle*. Two slow nodes skipping complementary slots could intersect rarely; slot 0 is polled simultaneously by everyone so in practice it's decided there, but it's an untested path.

## Something that helped me review this

I wrote an offline "slot alignment" table that rebuilds the grid from the manifest and classifies each `(slot, host)` as `ok` / `late` / `missed:<cause>` / `unaccounted` — it's what surfaced #2. On the 1.2× probe session under the late-fire rule it reads:

```
slot grid: start=…174.141 (recorded) interval=0.05s slots=15 hosts=2 skew_reference=node-a
  host    rows  late  missed      unacc  max_offset  max_skew
  node-a    15     0  -               0       0.01s     0.00s
  node-b    12     8  overrun=3       0       0.08s     0.10s

     seq   sched(+s)  node-a            node-b
       0        0.00  ok +0.01s         ok +0.04s
       1        0.05  ok +0.01s         late +0.06s
       2        0.10  ok +0.01s         late +0.08s
       3        0.15  ok +0.01s         missed:overrun
       4        0.20  ok +0.01s         ok +0.04s
       …
```

`unacc` > 0 on any host is the signature of #2 — a slot with neither a row nor a missed range. It needs the grid's wall-clock anchor in the manifest (`started_at_unix` is session construction, ~30 s before slot 0 on real runs), so it adds one field, `slot_grid_started_at_unix`, in `_start_collector`; older manifests fall back to a median-residual calibration and the header says so.

It's the second commit on the same fork branch, [`57c0dc5e`](https://github.com/FrankD412/srt-slurm/commit/57c0dc5e): [`slot_alignment.py`](https://github.com/FrankD412/srt-slurm/blob/57c0dc5e/src/srtctl/core/power/slot_alignment.py) (module), [`--slot-table`](https://github.com/FrankD412/srt-slurm/blob/57c0dc5e/src/srtctl/cli/validate_power_artifacts.py) on `srtctl-validate-power`, [tests](https://github.com/FrankD412/srt-slurm/blob/57c0dc5e/tests/test_power_slot_alignment.py). Separate from this PR — it doesn't touch the verdict — but happy to open it as a follow-up if it'd help the InferenceX triage.

FrankD412:
Pulled `104f4bcd` and ran the six touched suites (512 pass). #1–#4 and the small items all landed; the shared shutdown bracket is the shape I had in mind, and `test_final_scrape_uses_one_shared_slot_after_an_endpoint_hangs` pins it well.

One gap in the bracket path, found while checking coverage: the new `if shutdown_bracket is None` fallback ([L524-526](https://github.com/NVIDIA/srt-slurm/blob/104f4bcdc5863b35507ad0315da2d5630dc0fe37/src/srtctl/core/power/session.py#L524-L526)) is reachable, and when it is, it reintroduces the stale-counter bracket.

## Crash path skips the bracket

`_stop` is set from three places. `stop_and_finalize` ([L594-604](https://github.com/NVIDIA/srt-slurm/blob/104f4bcdc5863b35507ad0315da2d5630dc0fe37/src/srtctl/core/power/session.py#L594-L604)) arms `_shutdown_bracket` first; the endpoint-thread `except` ([L553](https://github.com/NVIDIA/srt-slurm/blob/104f4bcdc5863b35507ad0315da2d5630dc0fe37/src/srtctl/core/power/session.py#L553)) and the supervisor `except` ([L564](https://github.com/NVIDIA/srt-slurm/blob/104f4bcdc5863b35507ad0315da2d5630dc0fe37/src/srtctl/core/power/session.py#L564)) set it directly. So after a sibling crash every surviving thread leaves its loop with `_shutdown_bracket is None` and brackets at its own counter with `scheduled_at_unix=time.time()`. A survivor that was mid-request when the crash happened closes several slots behind its peers and files no missed range for the slots that elapsed during the hang — the same evidence hole as before, now confined to the crash path.

Reproduced on your head: node-b hangs on slot 2, node-a raises on slot 6 (crash slot ≈ 10). node-b brackets at **seq 3**, `missed_sample_ranges` for node-b is empty, `_shutdown_bracket` is `None` throughout.

Same scenario on a 1 s grid, before and after (red-filled bar = the request whose thread raised; it sets `_stop` and takes no bracket poll itself):

crash-path bracket: 104f4bcd leaves node-b at seq 3 with slots 3-6 unaccounted; with _request_stop node-b files 3-6 and brackets at B=7

<sup>Generator: [`review-assets/pr-480/timing_diagram.py`](https://github.com/FrankD412/srt-slurm/blob/6fafad851b2f41eacfe80e1bd46006c04f21b5fe/pr-480/timing_diagram.py) (`crash` scenario).</sup>

`publication_valid` is unaffected — `COLLECTOR_EXCEPTION` is fatal, so these runs were never publishable. This is about the manifest telling the truth on exactly the runs where someone will be reading it to find out what went wrong.

## Suggested fix

Route all three `_stop.set()` sites through one helper that arms the bracket first:

```python
def _request_stop(self, stopped_monotonic: float) -> None:
    with self._state_lock:
        if self._collector_grid is not None and self._shutdown_bracket is None:
            ...  # the existing L594-603 body
    self._stop.set()
```

Then the `None` fallback at L524 is unreachable and can raise instead of silently degrading — any future stop path that bypasses the helper fails a test rather than producing a misaligned manifest.

On my fork as one commit on top of your head, with the test: [FrankD412/srt-slurm@22ef40dc](https://github.com/FrankD412/srt-slurm/commit/22ef40dc) (branch [`pr-480-crash-bracket`](https://github.com/FrankD412/srt-slurm/tree/pr-480-crash-bracket), [compare](https://github.com/NVIDIA/srt-slurm/compare/104f4bcdc5863b35507ad0315da2d5630dc0fe37...FrankD412:srt-slurm:pr-480-crash-bracket)). `test_collector_crash_still_pins_every_survivor_to_one_bracket_slot` fails on `104f4bcd` (`_shutdown_bracket is None` after the crash) and passes with the change; 513 pass across the six suites; ruff/ty clean.

## Still open from the first pass

**#5** — the validation-policy relaxation (3 s → `interval + 2·timeout` = 5 s at defaults, up to 10 s on long windows) isn't mentioned in the PR body, and the eight InferenceX #3087 gap values aren't listed. No code ask; just the body and the numbers so a reader can see the knobs weren't fit to the data.

## Slot-alignment table

Rebased onto `104f4bcd` as [`slot-table-on-v3`](https://github.com/FrankD412/srt-slurm/tree/slot-table-on-v3) ([991fa671](https://github.com/FrankD412/srt-slurm/commit/991fa671)). Your range cap needed one adjustment on my side: no-row slots after the last retained range are now classified `missed:truncated` rather than `unaccounted` when `missed_sample_ranges_truncated` is set, so the cap doesn't read as lost accounting. Still a follow-up, not for this PR.

PR #481, opened 2026-09-19
fix(power): DCGM CPU power is the CPU rail; publish rail + SysIO breakdown, prefer ACPI in auto

FrankD412:
## Summary

DCGM-mode CPU power has been reporting the **CPU rail, not the socket envelope** — about half of what ACPI-mode reports for the same hardware (~53 W vs ~100 W per socket on GB200 reference runs). This PR makes that explicit in the data, adds the SysIO rail DCGM also exposes, and stops `auto` mode from picking DCGM over the fuller ACPI source.

Follows #453 (one row per socket, rail names in one module). Independent of #475 (Pareto/HTML report), which already renders the `warnings` this PR adds.

## Why (verified against NVIDIA/DCGM source)

DCGM has no CPU power backend of its own. The sysmon module walks `/sys/class/hwmon/*/device/power1_oem_info`, matches the label prefix, and reads `power1_average` from that channel — i.e. it is a second consumer of the same ACPI `power_meter` interface our ACPI reader uses. Pinned to `master` @ [`64df9f8`](https://github.com/NVIDIA/DCGM/tree/64df9f894541e426e416131a9820cae97aa4dd81):

| DCGM field | reads hwmon label | our rail |
|---|---|---|
| 1130 `DCGM_FI_DEV_CPU_POWER_WATTS` ([def](https://github.com/NVIDIA/DCGM/blob/64df9f894541e426e416131a9820cae97aa4dd81/dcgmlib/dcgm_fields.h#L2888), [getter](https://github.com/NVIDIA/DCGM/blob/64df9f894541e426e416131a9820cae97aa4dd81/modules/sysmon/DcgmModuleSysmon.cpp#L1336-L1341)) | `CPU Power Socket N` → `power1_average` ([L49-53](https://github.com/NVIDIA/DCGM/blob/64df9f894541e426e416131a9820cae97aa4dd81/modules/sysmon/DcgmSystemMonitor.cpp#L49-L53), [L165-168](https://github.com/NVIDIA/DCGM/blob/64df9f894541e426e416131a9820cae97aa4dd81/modules/sysmon/DcgmSystemMonitor.cpp#L165-L168)) | `cpu_rail` |
| 1132 `DCGM_FI_DEV_SYSIO_POWER_UTIL_CURRENT` ([def](https://github.com/NVIDIA/DCGM/blob/64df9f894541e426e416131a9820cae97aa4dd81/dcgmlib/dcgm_fields.h#L2898)) | `SysIO Power Socket N` → `power1_average` | `soc` |
| 1131 `DCGM_FI_DEV_CPU_POWER_LIMIT_WATTS` | `Grace Power Socket N` → `power1_cap` **only** ([L67-71](https://github.com/NVIDIA/DCGM/blob/64df9f894541e426e416131a9820cae97aa4dd81/modules/sysmon/DcgmSystemMonitor.cpp#L67-L71), [L183-186](https://github.com/NVIDIA/DCGM/blob/64df9f894541e426e416131a9820cae97aa4dd81/modules/sysmon/DcgmSystemMonitor.cpp#L183-L186)) | (envelope's cap, not its draw) |

No DCGM field reports the `Grace Power Socket N` envelope's average. Real samples (per-socket means over the whole run):

| run | source | channel | mean W |
|---|---|---|---|
| 3054237 | ACPI | `CPU Power` (cpu_rail) | 52.9 |
| 3054237 | ACPI | `Grace Power` (envelope, `power_w`) | **100.6** |
| 2998956 | ACPI | `CPU Power` / `Grace Power` | 55.4 / **104.0** |
| 430661 | DCGM | field 1130 (`power_w`) | 56.2 |

## What changes

**Vocabulary** — `cpu_rails.DCGM_FIELD_RAIL_KINDS = {1130: cpu_rail, 1132: soc}` (+ `DCGM_PRIMARY_FIELD_ID`, field names, hwmon labels) in the one module that owns rail names. `cpu_sample.dcgm_rail_readings()` is the one place field ids become rails: 1130 → the `dcgm` primary (`power_w`, unchanged) **and** the `cpu_rail` component; 1132 → `soc`. Host reader and scrape parser both call it.

**Producers**
- Python host collector (`srtctl.core.cpu_power`): watches 1130 + 1132; wide CSV gets `cpu_rail_w` = 1130 and `soc_w` = 1132; node metadata gains a `power_fields` table (field id, name, rail, hwmon label, column).
- Rust `cpu-power-exporter`: field group `[1130, 1132]`; `/metrics` emits one `cpu_power_dcgm_watts{socket,field_id,source}` sample per field, 1130 first so a legacy collector still meets it first. **`--source auto` is now ACPI-first** (was DCGM-first, unlike the Python collector) — that asymmetry is why run 430661 came out DCGM-mode on nodes whose hwmon almost certainly had the envelope.
- `cpu_parser`: reads the `field_id` label; an unlabelled sample (older exporter binary) is 1130, so old binaries keep working and now also get `cpu_rail_w` filled.

**Report** — `CpuSamples.sources`; a DCGM-sourced run gets the warning `CPU power source is DCGM field 1130 = CPU rail only (not the socket envelope; roughly half of ACPI-mode CPU power): not comparable with ACPI-mode runs`.

**Docs** — `cpu-power-telemetry.md` (field→rail table, `auto` order, column semantics), `config-reference.md`, `dcgm-4.7-runtime-support.md`, `CpuPowerExporterConfig` docstring; stale "rails blank for DCGM" comments corrected.

## Data correctness

- `power_w` / `total_power_w` semantics are **unchanged** for every mode: 1130 stays the DCGM-mode socket power (SysIO is never summed in), ACPI is untouched. Summing 1130+1132 into `power_w` was considered and rejected — the ACPI envelope is ~100 W vs cpu_rail+soc ≈ 59 W, so the sum would still be wrong while looking less wrong.
- Real-sample check: `power_energy_report --json-out` for **430661** (DCGM) and **3054237** (ACPI) is identical to the pre-change output except for the one new warning on 430661.
- A new-shape scrape body for 430661's first sample parses to `socket 0: power_w=50.149, rails={cpu_rail: 50.149, soc: 6.2}`, `total_power_w=96.914` — the same total the run recorded.

Compatibility:

| srtctl | exporter binary | result |
|---|---|---|
| new | new | `power_w`=1130, `cpu_rail_w`=1130, `soc_w`=1132 |
| new | old | `power_w`=1130, `cpu_rail_w`=1130, `soc_w` blank |
| old | new | as before (old parser ignores `field_id`, dedups by socket) |

## Testing

- `uv run pytest` on the power suites: 482 passed (cpu_rails / cpu_sample / cpu_parser / cpu_power / collector / contract / samples / energy report / validator / artifacts / telemetry). New cases: field-labelled body, body without 1130 (no socket published), garbage `field_id`, rail-only warning present for DCGM and absent for ACPI, host reader watching both fields with a socket missing 1132.
- `cargo test -p cpu-power-exporter`: 14 passed (`render_dcgm_metrics` ordering/labels, `POWER_FIELDS` pin). `cargo fmt --check` clean.
- `make lint`: ruff clean; `ty` diagnostic count unchanged (38, pre-existing).

## Notes for review

- Field 1132's semantics are verified from DCGM source; I have not yet confirmed it returns non-zero on a live GB200 node (`dcgmi dmon -e 1130,1132 -i cpu:0 -c 3`). If it reads blank, `soc_w` is simply left empty — nothing else depends on it.
- The `field_id` label on the existing metric (rather than a new metric family) was chosen for wire compactness; an old collector sees two samples per socket and relies on 1130 being emitted first, which the exporter guarantees and a unit test pins.

Tracking: AIP-1397 (https://linear.app/nvidia/issue/AIP-1397)


FrankD412:
## Review — verified the DCGM claim against source, ran the suites

**Verdict: looks right; two things to tighten before un-drafting, one open question.**

### Verified
- **The central claim holds.** I read `DcgmSystemMonitor.cpp` @ `64df9f8` independently: `GetCpuSocketIdFromContents` maps `"CPU Power Socket "` → usage file (`m_cpuSocketToPowerUsagePath`, served by `GetCurrentCPUPowerUsage` → 1130), `"SysIO Power Socket "` → `m_sysioSocketToPowerUsagePath` (1132), and `"Grace Power Socket "` → **`DCGM_POWER_CAP_FILE`** only (`power1_cap`, served by `GetCurrentPowerCap` → 1131). There is also a `"Module Power Socket "` → `GetCurrentModulePowerUsage` path (field **1133** `DCGM_FI_DEV_MODULE_POWER_UTIL_CURRENT`) — see the open question below.
- **Sample data agrees.** Across the 5 lyris_pareto_2 ACPI runs + 3054237 + 3074937: `Grace Power` ≈ 98–109 W/socket, `CPU Power` ≈ 51–56 W, `SysIO` ≈ 6.1–6.6 W. 430661 (DCGM, `CPU0:cpuPowerUsageW`) ≈ 51/58 W — squarely the CPU-rail band. So the report's 791 W vs 1,448 W for the same P1×4+D6×4 topology is exactly this.
- **Tests:** 428 passed across the cpu/power suites in a clean worktree. `ruff check` on the PR's files: clean (the 50 findings in `src tests` are all pre-existing on `upstream/main`, none in touched files). Couldn't run `cargo test` here (macOS 27 SDK / tapi linker breakage on this box, unrelated to the PR).
- `power_w` semantics unchanged for both modes; rejecting `1130+1132` as a fake envelope is the right call.

### Tighten
1. **Legacy-collector compatibility depends on scrape ordering that the parser doesn't actually assume.** The PR text says an old collector "relies on 1130 being emitted first" — true for the *old* parser, and the Rust test pins it. But the *new* `_parse_dcgm` uses `setdefault` per `(socket, field_id)`, so it's order-independent; fine. Worth stating in `cpu_parser.py` that the ordering guarantee exists only for old binaries, so nobody later "cleans up" the emission order thinking it's free.
2. **Warning text is per-run but the comparability problem is per-rollup.** `DCGM_CPU_RAIL_ONLY_WARNING` fires on a DCGM run in isolation, where it's informational; the real hazard is a combined report mixing ACPI and DCGM runs (the gb300 rollup does exactly this: 430661 alongside 3090373). Suggest also emitting a report-level note when `sources` differ across included runs — cheap to add in `build_combined_report` once #475 lands, or as a follow-up; noting it here so it isn't lost.

### Open question
- **Field 1133 (`Module Power Socket N`).** DCGM *does* have a usage getter for it. On GB200/GB300 the "module" is Grace+Blackwell superchip power, so it's almost certainly *not* the Grace envelope either and would double-count GPUs — but the PR's "no DCGM field reports the envelope" statement is only strictly true if 1133 is confirmed not to be it. A `dcgmi dmon -e 1130,1131,1132,1133 -i cpu:0 -c 3` on a live node alongside `cat /sys/class/hwmon/*/device/power1_oem_info` would settle both this and the "1132 non-zero" item you already flagged, and is worth pasting into the PR.

### Nits
- `cpu_rails.DCGM_FIELD_NAMES` / `DCGM_FIELD_HWMON_LABELS` / `DCGM_FIELD_RAIL_KINDS` are three parallel dicts keyed by field id; a single `dict[int, DcgmField(name, label, kind)]` would make it impossible for them to drift. Not blocking.
- `CpuSamples.sources: frozenset[str]` defaults to empty for fixtures, which means a fixture-built DCGM sample never triggers the warning — fine, but a one-line comment in the test that asserts the warning saying it sets `source=` explicitly for that reason would help the next reader.

FrankD412:
Thanks — addressed in e283ec99.

**Done**
- **Tighten 1 (ordering):** comment added in `cpu_parser._parse_dcgm` stating the new parser keys on `(socket, field_id)` and is order-independent; the exporter's "1130 first per socket" order is kept solely for older srtctl parsers and shouldn't be cleaned up while old collectors may scrape new binaries. `test_cpu_rails` pins `DCGM_POWER_FIELD_IDS == (1130, 1132)` with that rationale.
- **Nit (parallel dicts):** collapsed into one `DcgmPowerField(field_id, name, hwmon_label, kind)` record per field — `DCGM_POWER_FIELDS` / `DCGM_FIELD_BY_ID`; the `DCGM_FIELD_RAIL_KINDS` / `DCGM_POWER_FIELD_IDS` views consumers use are derived from it. Metadata `power_fields` now iterates the records.
- **Nit (fixture `sources`):** comment on the warning test explaining it loads through the CSV because `CpuSamples.sources` is populated from the `source` column and direct fixtures leave it empty.
- **Codecov line:** the `entity_id not in self._cpu_ids` guard in `DcgmCpuPowerReader.read_watts` is now exercised (a value for an un-enumerated entity is dropped).
- **Open question (1133):** you're right that "no DCGM field reports the envelope" was only strictly true with 1133 ruled out. I can't run the live check from here, so the branch now says exactly what is known: `cpu_rails` documents 1131 (cap) and 1133 (`Module Power Socket N`, scope on GB200/GB300 unverified) as deliberately excluded, `test_cpu_rails` asserts neither is in the table, and the docs table lists 1133 as unverified/excluded with the `dcgmi dmon -e 1130,1131,1132,1133 -i cpu:0 -c 3` + `cat /sys/class/hwmon/*/device/power1_oem_info` recipe. Will paste the live output here once someone has a node.

**Deferred**
- **Tighten 2 (rollup-level mixed-source warning):** agreed it's the real hazard (430661 next to 3090373 in the gb300 rollup). `build_combined_report` lives in #475, so this goes there once it lands — tracked as a checkbox on AIP-1397 rather than in this PR.

Verification unchanged: report JSON for 430661 and 3054237 identical to the pre-change output except the warning; 132 power tests + 14 cargo tests pass, ruff clean, `ty` count unchanged.

